PR #6, opened 2022-01-27
Solution

Krimat:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from ..player import Player
```
TPolina, line 1: it's better to use the full path to the module. the same in other files

TPolina (changes requested):

TPolina (approved):

MasterpieceElbow (approved):

PR #8, opened 2022-01-27
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list):
+    return sum([teammate.get_rating() for teammate in team])
```
cloudprogramm, line 2: It is not necessary write here a list generator. If u remove it code will still work ;)

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,19 @@
...
+    def eat_favourite_dish(self):
+        print(f"{self.nickname} is eating {self._favourite_dish}")
+
+    @abstractmethod
+    def get_rating(self):
+        pass
+
+    @abstractmethod
+    def player_info(self):
+        pass
```
cloudprogramm, line 19: Maybe it's a little chit but i don't wrote the last two functions with `@anstractmethod`. I define them in child classes and everithing works)

app/players/elves/elf.py
```diff
@@ -0,0 +1,20 @@
...
+        print(f"{self.nickname} is playing a song "
+              f"on the {self._musical_instrument}")
+
+    @abstractmethod
+    def get_rating(self):
+        pass
+
+    @abstractmethod
+    def player_info(self):
+        pass
```
cloudprogramm, line 20: The same cheat you can use)

cloudprogramm (approved):
Good job!

app/players/dwarves/dwarf.py
MasterpieceElbow: If you have an abstract class that inherits another abstract class you don't need to repeat abstract methods from parent.

app/main.py
MasterpieceElbow: Yes, it would be better to remove square brackets

app/players/elves/elf.py
MasterpieceElbow: The same as Dwarf

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #14, opened 2022-01-27
'Solution'

serhii56465:


app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Nazar-kob, line 4: Add an ABC class after the player, so that another programmer knows, that your class is inheriting an abstract class!

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Nazar-kob, line 4: same problem, but I approve

Nazar-kob (approved):

MasterpieceElbow (approved):

PR #16, opened 2022-01-27
Solution

nikon4uk:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.players.elves import *
+from app.players.dwarves import *
+
```
phantom-terminal, line 3: You don't need to import like this, it's bad practice. And actually, you may not use import in main.py

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.players.elves import *
+from app.players.dwarves import *
```
phantom-terminal, line 2: You don't need to import like this, it's bad practice. And actually, you may not use import in main.py

phantom-terminal (changes requested):

app/main.py
nikon4uk: thanks!

phantom-terminal (approved):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    @abstractmethod
+    def player_info(self):
+        pass
+
+    @abstractmethod
+    def get_rating(self):
+        pass
```
MasterpieceElbow, line 16: If you have an abstract class that inherits another abstract class you don't need to repeat abstract methods from parent.

app/players/elves/elf.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__(nickname)
+        self._musical_instrument = musical_instrument
+
+    @abstractmethod
+    def get_rating(self):
+        pass
+
+    @abstractmethod
+    def player_info(self):
+        pass
```
MasterpieceElbow, line 16: If you have an abstract class that inherits another abstract class you don't need to repeat abstract methods from parent.

MasterpieceElbow (changes requested):

IvanRamyk (approved):
Well done!

PR #18, opened 2022-01-27
Solution

phantom-terminal:


vldtsw (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(dream_team: list):
+    return sum((mate.get_rating() for mate in dream_team))
```
MasterpieceElbow, line 2: Get rig of extra brackets in `sum`

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,20 @@
...
+    def eat_favourite_dish(self):
+        print(f"{self.nickname} is eating {self._favourite_dish}")
+
+    @abstractmethod
+    def player_info(self):
+        pass
+
+    @abstractmethod
+    def get_rating(self):
+        pass
```
MasterpieceElbow, line 20: If you have an abstract class that inherits another abstract class you don't need to repeat abstract methods from parent.

app/players/elves/elf.py
```diff
@@ -0,0 +1,21 @@
...
+        print(f"{self.nickname} is playing a song on the"
+              f" {self._musical_instrument}")
+
+    @abstractmethod
+    def player_info(self):
+        pass
+
+    @abstractmethod
+    def get_rating(self):
+        pass
```
MasterpieceElbow, line 21: If you have an abstract class that inherits another abstract class you don't need to repeat abstract methods from parent.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #20, opened 2022-01-27
solution

NikitaZH1812:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Abnormaltype, line 4: You also need to make it clear that this is an abstract class. `class Dwarf(Player, ABC):`

Abnormaltype (changes requested):

app/players/dwarves/dwarf.py
NikitaZH1812: thx, fixed

Abnormaltype (approved):
Nice job!

MasterpieceElbow (approved):

PR #29, opened 2022-05-20
solution

MarkoKhodan:


app/main.py
```diff
@@ -1 +1,55 @@
...
+                 ),
+    DwarfWarrior(nickname="Dwarf",
+                 favourite_dish="Caesar Salad",
+                 hummer_level=3
+                 ),
+    DwarfBlacksmith(nickname="Smith",
+                    favourite_dish="Borsch",
+                    skill_level=4)
+]
+feast_of_the_dwarves(dwarves)
```
Abnormaltype, line 55: Do you need this?
$24 - $55

Abnormaltype (changes requested):

app/main.py
MarkoKhodan: No, I don't need it, but if I remove it, then I can't pass flake8, because got message that imports are not used

Abnormaltype: remove these imports?

Abnormaltype (approved):
Good job!

MarkoKhodan:
Thanks

PR #32, opened 2022-05-23
Solution

YuriiKindrat:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player, ABC
+
+
+class Dwarf(Player, ABC):
```
rotsen18, line 4: You can select inheritance only from Player, because Player class already inherit from ABC.
Also delete from import ABC.
So you should fix this issue in all files

rotsen18 (changes requested):

app/players/dwarves/dwarf.py
Magus6784: If we do this way we have warning about "Where is @abstractmethod?"

Magus6784 (approved):

rotsen18 (approved):

MasterpieceElbow (approved):

PR #35, opened 2022-05-24
Solution

Magus6784:


app/players/elves/elf_ranger.py
```diff
@@ -0,0 +1,16 @@
+from app.players.elves.elf import Elf
+
+
+class ElfRanger(Elf):
+
+    def __init__(self, nickname: str, musical_instrument: str,
+                 bow_level: int):
```
DmytroSvirsa, line 7: I think if we breach the argument line, it's better to have each argument from a new line for better readability, but your style has no conflicts with pep8

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+
```
DmytroSvirsa, line 1: looks like not needed blank line

DmytroSvirsa (approved):

app/players/elves/elf_ranger.py
Magus6784: My point is - I write argument who go in super in one line
And new argument in another line, so we can see "new" argument for this class and "old" argument which go to "father"

MasterpieceElbow (approved):

PR #36, opened 2022-05-24
Solution

rotsen18:


app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Flashmobber, line 4: If you inherit this class also from the ABC class, you will get rid of the message from PyCharm like "Class Elf must implement all abstract methods". It's just a small advice. It's up to you to accept it or not.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Flashmobber, line 4: and same here

Flashmobber (approved):

MasterpieceElbow (approved):

PR #38, opened 2022-05-24
'Solution'

Korenely:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self, nickname: str):
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self):
+        pass
+
+    @abstractmethod
+    def player_info(self):
+        pass
```
Flashmobber, line 14: Please, add a type annotation to all method in all classes and all modules. It's a good practice.

Flashmobber (changes requested):

MasterpieceElbow (approved):

PR #40, opened 2022-05-24
Solution

4ndyua:


app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, nickname):
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self):
+        pass
+
+    @abstractmethod
+    def player_info(self):
+        pass
```
Flashmobber, line 15: Please, add a type annotation to all method in all classes and all modules. It's a good practice.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+
```
Flashmobber, line 2: This is redundant empty string, it is better to delete it.

Flashmobber (changes requested):

app/players/player.py
4ndyua: Fixed

app/players/dwarves/dwarf.py
4ndyua: Fixed

MasterpieceElbow (approved):

PR #48, opened 2022-05-25
'Solution'

daria-niklaevna:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team):
+    return sum([i.get_rating() for i in team])
+
+
+def elves_concert(elves):
+    for i in elves:
```
Chardona, line 6: You should use more infomative names, than "i". It's still solid code but can be hard for understanding to the others

app/main.py
```diff
@@ -1 +1,12 @@
...
+        i.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves):
+    for i in dwarves:
```
Chardona, line 11: same

Chardona (approved):
Neat

app/main.py
```diff
@@ -1 +1,12 @@
...
+        i.eat_favourite_dish()
```
MasterpieceElbow, line 12: `i` is not a good variable name. Refactor it everywhere.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #59, opened 2022-05-25
Solution

Kiawami:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
```
Hectorovich, line 1: You don`t need abc module here. It is necessary only in module "Player"

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Hectorovich, line 5: And Dwarf must inherit only class Player.

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
```
Hectorovich, line 1: Here too.

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
Hectorovich, line 5: And here.

Hectorovich (changes requested):
And also you don`t need packages in this task. Just use directories instead. To fix it you can just delete "__init__.py" files from packages and after this your packages will transform to directories.

app/players/elves/elf.py
Kiawami: well I just put ABC there because of that https://prnt.sc/rqiCg0VS0CZL
and in the task Elf and Dwarf were also mentioned to be abstract, so..

Hectorovich (approved):

MasterpieceElbow (approved):

PR #62, opened 2022-05-26
Solution

lenasharabura:


NOSSUDAK (approved):
Why you don't use type annotation?

PR #64, opened 2022-05-26
Solution

Igor-Cegelnyk:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list):
+    return sum([player.get_rating() for player in team])
```
IvanRamyk, line 2: You can prevent creating a redundant list via passing comprehension without square brackets

IvanRamyk (approved):
Well done!

PR #65, opened 2022-05-26
Solution

mariiia-movchan96:


rotsen18 (approved):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def calculate_team_total_rating(players: list):
+    return sum([player.get_rating() for player in players])
+
+
+def elves_concert(elves: list):
+    return [elf.play_elf_song() for elf in elves]
```
TPolina, line 6: you should not return anything here. `play_elf_song()` returns `None`, so you just return a list of `None`s

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+
+def feast_of_the_dwarves(dwarves: list):
+    return [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
TPolina, line 10: same here

TPolina (changes requested):

PR #70, opened 2022-06-07
Solution

drivemax26:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team):
+    return sum([member.get_rating() for member in team])
```
IvanRamyk, line 2:
```suggestion
    return sum(member.get_rating() for member in team)
```
When you write the sum like this you avoid creating a new list

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #121, opened 2022-09-14
My dwarves and Elves

SorokaIgor:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list):
+    calc_rate = 0
+    for player in players:
+        calc_rate += player.get_rating()
+    return calc_rate + 3
```
Abnormaltype, line 5: why did you add 3?

Abnormaltype (changes requested):

app/main.py
SorokaIgor: image

i saw that at the example.

PR #132, opened 2022-09-20
'Solution'

AVyha:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(list_of_players: list) -> int:
+    return sum([player.get_rating() for player in list_of_players])
+
+
+def elves_concert(list_of_elfs: list) -> None:
```
mltdll, line 5: I don't think it's a reason to request changes, but you have spelling issues in **elfs** and **dwarfes**: there are examples of right spellings in the names of the functions. :)

mltdll (approved):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for i in range(len(list_of_elfs)):
+        list_of_elfs[i].play_elf_song()
```
mltdll, line 7: You can also just use for loops over lists of elves and dwarves.

mltdll (changes requested):
Actually, I asked mentors if I should request changes in this case, and they said yes, I should. Sorry!

app/main.py
```diff
@@ -2,11 +2,11 @@ def calculate_team_total_rating(list_of_players: list) -> int:
     return sum([player.get_rating() for player in list_of_players])
 
 
-def elves_concert(list_of_elfs: list) -> None:
-    for i in range(len(list_of_elfs)):
-        list_of_elfs[i].play_elf_song()
+def elves_concert(list_of_elves: list) -> None:
+    for elv in list_of_elves:
```
mltdll, line 6: Now you should change **elv** to **elf**. :)

AVyha:
> Actually, I asked mentors if I should request changes in this case, and they said yes, I should. Sorry!

Thank for the rewiev) Fix problem already)

app/main.py
AVyha: :)

mltdll (approved):
GJ! :)

app/main.py
AVyha: sorry for that) i'm very careless

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(list_of_players: list) -> int:
+    return sum([player.get_rating() for player in list_of_players])
```
Y-Havryliv, line 2: get rid of unnecessary `[]` here

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #134, opened 2022-09-20
Solution

mltdll:


AVyha (approved):
all good, and you don't have problem with elf and dwarves like me)

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABCMeta, abstractmethod
+
+from app.players.player import Player
+
+
+# metaclass=ABCMeta makes IDE recognize that this is also an abstract class.
+class Dwarf(Player, metaclass=ABCMeta):
+    @abstractmethod
```
Y-Havryliv, line 8: why did you make `__init__` an abstract method?

Y-Havryliv: btw, you can inherit this class from `ABC` and avoid using `metaclass`

app/players/elves/elf.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABCMeta, abstractmethod
+
+from app.players.player import Player
+
+
+# metaclass=ABCMeta makes IDE recognize that this is also an abstract class.
+class Elf(Player, metaclass=ABCMeta):
+    @abstractmethod
```
Y-Havryliv, line 8: `__init__` shouldn't be an abstract method

Y-Havryliv (changes requested):

app/players/dwarves/dwarf.py
mltdll: I thought `__init__` should be abstract because it was required to implement in all the subclasses. Can you please clarify why it shouldn't?

Nattalli (approved):
Well done! 👍🏻

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    @abstractmethod
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
```
Nattalli, line 7: but actually - remove this method at all
Every class in Python has implemented `__init__` method. I think that it is not needed here

Nattalli (changes requested):

app/players/player.py
mltdll: Ok, I did. But there's this line in the README...
> The base class should be the `Player` class, which should have a single `nickname` attribute, where the player's name will be stored.

Nattalli: SORRY, I FORGOT TO REMOVE MY COMMENT
I understood that your solution was right... Can you rollback the `__init__` method? 🥲

app/players/player.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    @abstractmethod
```
Nattalli, line 5: sorry, the `__init__` method should be here...

Nattalli (changes requested):

app/players/player.py
mltdll: Sure, no problem! :)

Nattalli (approved):
Good

PR #156, opened 2022-09-26
Solution

Yaroslav-Stetsko:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
+    total_rating = 0
```
Y-Havryliv, line 2: you can do it in a one line, but it's up to you

app/main.py
```diff
@@ -1 +1,15 @@
...
+    return total_rating
+
+
+def elves_concert(elves: list) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
+    for Dwarf in dwarves:
```
Y-Havryliv, line 14: `for dwarf in dwarves:` maybe?

Y-Havryliv (changes requested):

app/main.py
Yaroslav-Stetsko: Fixed

app/main.py
Yaroslav-Stetsko: Fixed

Y-Havryliv (approved):
GJ!

PR #166, opened 2022-09-27
Created classes from abstract class elves and others

alexzubkoff:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
```
serviza, line 3: here you don't need to import classes

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+from abc import ABC
```
serviza, line 2: it is not necessary

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+from abc import ABC
```
serviza, line 2: the same

serviza (changes requested):

app/main.py
alexzubkoff: I will have an arrors, becuse i have they as arg annotations list[Player], list[Elf] ....

alexzubkoff: from abc import ABC is necessary!!! Code will be work without that. But today with Danya we discussed and he said that it is a good practice to write because Elf is abstract class and you can not remember it in big project. 
But when you see ABC,  you know that it is abstract. I think it is right.

app/players/elves/elf.py
alexzubkoff: from abc import ABC is necessary!!! Code will be work without that. But today with Danya we discussed and he said that it is a good practice to write because Elf is abstract class and you can not remember it in big project. 
But when you see ABC,  you know that it is abstract. I think it is right.

app/main.py
serviza: > I will have an arrors, because i have they as arg annotations list[Player], list[Elf] ....
Ok. I did not notice these annotations...

app/players/dwarves/dwarf.py
serviza: I see, you are more vigilance))

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def calculate_team_total_rating(players_list: list[Player]) -> int:
+    return sum([player.get_rating() for player in players_list])
```
Nattalli, line 7: get rid of unnecessary `[]` here

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def elves_concert(elf_list: list[Elf]) -> None:
+    for elf in elf_list:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarf_list: list[Dwarf]) -> None:
+    for dwarf in dwarf_list:
```
Nattalli, line 16: dwarf in plural is `dwarves`

Nattalli (changes requested):

app/main.py
alexzubkoff: Ok, thanks! And i miss elf in plural too. Now i will change.

Nattalli (approved):
Looks good 😄

PR #169, opened 2022-09-27
'Solution'

OvDima1:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
```
LiudmylaKulinchenko, line 3: You don't have to import all these classes

LiudmylaKulinchenko: Probably than your annotation won't work. But I'm not sure that 3 imports worth it

app/players/elves/druid.py
```diff
@@ -0,0 +1,18 @@
+from app.players.elves.elf import Elf
+
+
+class Druid(Elf):
+    def __init__(self,
+                 nickname: str,
+                 musical_instrument: str,
+                 favourite_spell: str):
```
LiudmylaKulinchenko, line 8: Forgot annotation here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
looks good for me, nice job

Nattalli (approved):
Good!

PR #174, opened 2022-09-27
Finished the task

Cerne13:


VadymShkarbul (approved):
Good

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    def eat_favourite_dish(self) -> None:
+        print(f'{self.nickname} is eating {self._favourite_dish}')
```
MasterpieceElbow, line 10: Use `"` instead of `'` by default.

MasterpieceElbow (changes requested):

app/players/dwarves/dwarf.py
VadymShkarbul: I don't agree. If it is necessary to use such a construction, it will be more convenient to use single ones first: f'"{bourbon_name}" is the best bourbon ever'

Cerne13: I also agree with Vadym. Naming with double quotes inside an f-string is more readable than backslash-blocking because of double-quote usage for the string notation itself.
But ok, if the current code style demands...

MasterpieceElbow: @VadymShkarbul @Cerne13  You are right, but here you don't have `"` inside the string.

MasterpieceElbow (approved):

PR #178, opened 2022-09-27
Solution

VadymShkarbul:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
...
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    def __init__(self, nickname: str, favourite_dish: str):
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    @abstractmethod
+    def eat_favourite_dish(self):
```
FranchukS, line 11: Method "eat_favourite_dish" in Dwarf class is the same for both subclasses so you can use it here once

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
...
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    def __init__(self, nickname: str, musical_instrument: str):
+        super().__init__(nickname)
+        self._musical_instrument = musical_instrument
+
+    @abstractmethod
+    def play_elf_song(self):
```
FranchukS, line 11: Method "play_elf_song" in Elf class is the same for both subclasses so you can use it here once

FranchukS (changes requested):

FranchukS (approved):
Well done!

Nattalli (approved):
Good job!

PR #185, opened 2022-09-28
Solution

mariia-movchan:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list):
+    return sum(player.get_rating() for player in players)
+
```
FranchukS, line 3: Add type annotation for return value in main.py

FranchukS (changes requested):

app/main.py
mariia-movchan: Thx, fixed :)

FranchukS (approved):
Well done!

MasterpieceElbow (approved):
GJ!

PR #187, opened 2022-09-28
Dwarves and Elves

VitaliyBychevoy:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list) -> int:
+    team_list = []
+    for player in players:
```
FranchukS, line 7: You can use list comprehension here, but it's optional

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
```
FranchukS, line 2: You don`t have to import these modules, your code will work without them.

FranchukS (changes requested):

app/main.py
VitaliyBychevoy: I have becouse class Dwarf  and class Elf were appied in  def elves_concert, def feast_of_the_dwarves

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,20 @@
...
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
+               f"with skill of the {self._skill_level} level"
```
MasterpieceElbow, line 17: Avoid using `\`. Use `( )`. Remake everywhere.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #200, opened 2022-10-12
Solution

Terrrya:
There is a mistake in Readme:

_elves = [
    Druid(nickname="Nardual", musical_instrument="flute", favourite_spell="aaa"),
    ElfRanger(nickname="Rothilion", musical_instrument="trumpet", bow_level=33),
**],**_

_"elves_concert - it should take a list of Elf"_ - it is says, but in the example there is a comma at the end. So elves is tuple in example : _elves = [...],_

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    return sum(player.get_rating() for player in players)
+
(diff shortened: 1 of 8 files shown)


IvanRamyk (approved):
Good job!
Thanks for feedback, readme is corrected :)

PR #213, opened 2022-11-02
Solution

darynka-h:
Solution

### DIFF
--- a/app/players/dwarves/dwarf.py
+++ b/app/players/dwarves/dwarf.py
@@ -1,0 +1,12 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    @abstractmethod
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    def eat_favourite_dish(self) -> None:
+        print(f"{self.nickname} is eating {self._favourite_dish}")
--- a/app/players/elves/elf.py
+++ b/app/players/elves/elf.py
@@ -1,0 +1,13 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    @abstractmethod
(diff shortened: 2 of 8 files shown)


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    @abstractmethod
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
```
Serhii-Leonenko, line 7: Are you sure that `__init__` is abstractmethod?)
The same in other similar cases

app/players/player.py
```diff
@@ -0,0 +1,13 @@
...
+class Player(ABC):
+    @abstractmethod
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    def get_rating(self) -> None:
+        pass
+
+    def player_info(self) -> None:
+        pass
```
Serhii-Leonenko, line 13: I think these two methods `get_rating player_info` should be abstract)

Serhii-Leonenko (changes requested):

app/players/player.py
darynka-h: There are 2 method in Player class. But in Dwarf class and Elf class I can't define it, because I don't have any condition to do it. Of course I could be wrong, but I don't know how to fix it =(

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    @abstractmethod
```
IvanRamyk, line 6: should be fixed

app/players/player.py
IvanRamyk: you can leave them abstract
actually should define an abstract method in a child class only if it's non-abstract. `Efl` and `Dwarf` are abstract classes so you can leave some of the parent's methods non-implemented

IvanRamyk (changes requested):

PR #226, opened 2022-11-17
Solution

Vladyslav7125:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(list_of_player: list) -> int:
```
Abnormaltype, line 1: add type annotation

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in list_of_player)
+
+
+def elves_concert(list_of_elves: list) -> None:
```
Abnormaltype, line 5: add type annotation

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in list_of_elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(list_of_dwarves: list) -> None:
```
Abnormaltype, line 10: add type annotation

Abnormaltype (changes requested):

app/main.py
Vladyslav7125: fix

app/main.py
Vladyslav7125: fix

app/main.py
Vladyslav7125: fix

PR #246, opened 2022-12-06
solution

kleotan901:


app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
YuriiKindrat, line 5: No need to inherit from ABC class here

Abnormaltype (approved):

PR #256, opened 2022-12-14
Solution

sorrow03:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
katevasilenko, line 5: I don’t think you need to inherit from ABC here

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Elf(Player, ABC):
```
katevasilenko, line 5: And here too

katevasilenko (changes requested):
Good solution, just need small fix

app/players/elves/elf.py
sorrow03: i saw that in checklist and decided to do that

IvanRamyk (approved):
Well done!

PR #258, opened 2022-12-20
Solution

alina-boichenko:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```
katevasilenko, line 5: You don’t need to inherit from ABC here

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
katevasilenko, line 5: You don’t need ABC inheritance here too

katevasilenko (changes requested):
Please remove unnecessary inheritance. Everything else is good

IvanRamyk (approved):

PR #260, opened 2022-12-22
Solution

DmytroProkofiev:


app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elves: list) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarfs: list) -> None:
+    for dwarf in dwarfs:
+        dwarf.eat_favourite_dish()
```
Abnormaltype, line 15: add more descriptive type annotations

Abnormaltype (changes requested):

app/main.py
DmytroProkofiev: What's exactly means? For **pleyers, elves, dwarfs** "list" annotation is correct. The 1-st function returns an integer, 2-nd and 3-rd - nothing. Or for 2-nd and 3-rd functions "str" could be better annotation?

YuriiKindrat: This means that u should add smth like this list[<type inside list>] - more descriptivve type annotations

YuriiKindrat: also players - not pleyers

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
andrewbelo, line 1: Анотації треба розширити
Типу list чого? list[Dwarfs] і тд

andrewbelo (changes requested):

Serhii-Leonenko (approved):
GJ!

PR #264, opened 2022-12-29
Solution

Vitalii-Khmura:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
this class also should inherit from the abc class

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
YuriiKindrat, line 4: the same is for this class

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #273, opened 2023-01-04
Solution

yuliia-stopkyna:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
katevasilenko, line 1: Could you provide more detailed annotation please? List if what?

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list) -> None:
```
katevasilenko, line 5: The same

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
katevasilenko, line 10: And here

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
katevasilenko, line 5: You don’t need to inherit from `ABC` as `Player` already does it

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
katevasilenko, line 5: The same here with `ABC` inheritance

katevasilenko (changes requested):

katevasilenko (approved):
Good job!

PR #283, opened 2023-01-16
Solution

serg121192:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
olena-hul, line 1:
```suggestion
def calculate_team_total_rating(players: list[Player]) -> int:
```
Fix list annotation in other places

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,22 @@
...
+                 nickname: str,
+                 favourite_dish: str,
+                 skill_level: int
+                 ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
+               f"with skill of the {self._skill_level} level"
```
olena-hul, line 19: Use `()` instead of `\` to split a line

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
```
olena-hul, line 1: Is it really needed here?

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,22 @@
...
+from app.players.dwarves.dwarf import Dwarf
+
+
+class DwarfBlacksmith(Dwarf):
+
+    def __init__(self,
+                 nickname: str,
+                 favourite_dish: str,
+                 skill_level: int
+                 ) -> None:
```
olena-hul, line 13:
```suggestion
    def __init__(
        self,
        nickname: str,
        favourite_dish: str,
        skill_level: int
    ) -> None:

```

olena-hul: Fix in other places

app/players/dwarves/dwarf_blacksmith.py
olena-hul: Fix in other places

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
olena-hul, line 7: No need to inherit additionally from ABC. If the class does not implement abstact methods - it will remain abstract

olena-hul (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
ArsenPidhoretskyi, line 14: Must return string as I see

ArsenPidhoretskyi (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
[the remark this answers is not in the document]
serg121192: > It must return int varialbe

I don't understand what's wrong with my code. If we're talking on Player-class level, I guess None is right annotation for function. Because, if we're talking about other classes level, a could re-implement those functions (get_rating and player_info) with other functionality for each child class, that's why the annotations could be different. And I implemented that on other classes levels. That's why I don't understand what's wrong?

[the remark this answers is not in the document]
ArsenPidhoretskyi: For example `calculate_team_total_rating` receive a `list[Player]` where method get_rating returns None. How do you understand that we can use the sum function(if we do sum([None, None]) we will receive an error? We need to provide a base type for all abstract methods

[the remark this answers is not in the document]
olena-hul: @ArsenPidhoretskyi Good point. @serg121192  Please make the changes

olena-hul (approved):

olena-hul:
Good job!

PR #288, opened 2023-01-18
Solution

NataliiaPodoroha:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```
olena-hul, line 5: No need to inherit from ABC. Class that doesn't implement all abstract methods will remain abstract

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Elf(Player, ABC):
```
olena-hul, line 5: No need to inherit from ABC

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
olena-hul, line 9: Return annotation should be int

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
olena-hul, line 13: str here

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum([player.get_rating() for player in players])
```
ArsenPidhoretskyi, line 7: It is not required to do list comprehensions
```suggestion
    return sum(player.get_rating() for player in players)
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done)

PR #297, opened 2023-01-27
Solution

ilushasm:


app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
Vasyl-Smutok, line 6: no need to inheritance from `ABC` because `Player`  inheritance from `ABC`

Vasyl-Smutok (approved):
Good Job 👍

PR #304, opened 2023-02-03
Solution

leetwinoff:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from __future__ import annotations
+
+
+def calculate_team_total_rating(players: list) -> int:
```
ArsenPidhoretskyi, line 4: add full type hinting
list[Player]
The same in other functions. But with other type

app/main.py
```diff
@@ -1 +1,18 @@
...
+    total_rating = 0
+    for player in players:
+        total_rating += player.get_rating()
+    return total_rating
```
ArsenPidhoretskyi, line 8: use sum() function

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def get_rating(self) -> int:
+        return self._skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} with " \
+               f"skill of the {self._skill_level} level"
```
ArsenPidhoretskyi, line 19:
```suggestion
        return (
            f"Dwarf blacksmith {self.nickname} with "
            f"skill of the {self._skill_level} level"
        )
```

ArsenPidhoretskyi: If you can try to avoid usage of '\'

ArsenPidhoretskyi: If you fix here, fix in other places

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. We need to say the type which will be return on all child classes

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
ArsenPidhoretskyi, line 13: It returns str

ArsenPidhoretskyi (changes requested):

katevasilenko (approved):
Looks good to me

PR #310, opened 2023-02-07
'Solution'

natalie-goriela:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
olena-hul, line 9:
```suggestion
    def get_rating(self) -> int:

```

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
olena-hul, line 13:
```suggestion
    def player_info(self) -> str:

```

olena-hul (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Bohdan-Salamakha, line 6: u don't need to inherit from ABC one more time, u already did it because Player is inherited from ABC

app/players/elves/elf.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
Bohdan-Salamakha, line 6: u don't need to inherit from ABC one more time, u already did it because Player is inherited from ABC

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (approved):
Good job!

Bohdan-Salamakha (approved):

PR #314, opened 2023-02-08
Solution

bohdan-yatsyna:


app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def get_rating(self) -> int:
+        return self._skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} with " \
+               f"skill of the {self._skill_level} level"
```
Vasyl-Smutok, line 19: better to use () then /

Please fix it here and other places

```suggestion
        return (f"Dwarf blacksmith {self.nickname} with " 
                f"skill of the {self._skill_level} level")

```

Vasyl-Smutok (changes requested):

app/players/dwarves/dwarf_blacksmith.py
bohdan-yatsyna: Noted with thanks. And already fixed.

Vasyl-Smutok (approved):
Good Job 👍

PR #315, opened 2023-02-08
Solution

manjustice:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
Nattalli, line 1: better to clarify the type of instances inside of list

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elfs: list) -> None:
```
Nattalli, line 5: better to clarify the type of instances inside of list

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elfs:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
Nattalli, line 10: better to clarify the type of instances inside of list

Nattalli (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: List[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elfs: List[Elf]) -> None:
+    for elf in elfs:
```
Serhii-Leonenko, line 13: `elves` will be correct)

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Serhii-Leonenko, line 4: The class must inherit from the ABC also, the same about Elf

Serhii-Leonenko (changes requested):

app/players/dwarves/dwarf.py
manjustice: I don't understand why, because Dwarf and Elf already abstract classes, they are inherited from the Player class, which is inherited from ABC.

Serhii-Leonenko (approved):
Well done!

PR #318, opened 2023-02-09
Solution

bihunva:


app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,18 @@
...
+                 nickname: str,
+                 favourite_dish: str,
+                 skill_level: int) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        name = self.nickname
+        lvl = self._skill_level
+        return f"Dwarf blacksmith {name} with skill of the {lvl} level"
```
MaybeTI, line 15:
```suggestion
        return f"Dwarf blacksmith {self.nickname} with skill of the {self._skill_level} level"
```

MaybeTI: I think it's better, but you don't have to change

Nattalli: btw, agree

Nattalli (approved):
Well done!

PR #324, opened 2023-02-10
Solution

NEkropo1:


NEkropo1:
`def elves_concert(elves: list[Elf]) -> None:
    lul = [lul() for lul in map(lambda elf: elf.play_elf_song, elves)]  # Never ever do like this`
    
But it prints result and passes tests xD

app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_rating(self) -> int:
+        return self._hummer_level + 4
+
+
+warrior = DwarfWarrior(
+    nickname="Thiddeal",
+    favourite_dish="French Fries",
+    hummer_level=7
+)
```
DavidGlusch, line 28: redundant instance

DavidGlusch (changes requested):

app/players/dwarves/dwarf_warrior.py
Morty67: remove

Morty67 (changes requested):

app/players/dwarves/dwarf_warrior.py
NEkropo1: image
???

yura-hudzovskyi (approved):
Good job 🎆

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
fix this, please

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
YuriiKindrat, line 4: the same

YuriiKindrat (changes requested):

app/players/elves/elf.py
NEkropo1: image
According to definition in https://mate.academy/learn/python-core/python-core-oop-encapsulation-polymorphism-abstraction#/video/2908
Abstract class is class inherited from ABC which has at least one not overriden abstract method.
Since we inherit Dwarf and Elf from Player, who inherits from ABC and has 2 abstractmethods,
Elf and Dwarf are abstract too,
So, I implemented all 3 abstract classes.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+import abc
+from app.players.player import Player
+
+
+class Dwarf(Player, abc.ABC):
```
Serhii-Leonenko, line 5: Just import ABC class from abc module. The same in all other similar cases

Serhii-Leonenko (changes requested):

app/players/dwarves/dwarf.py
NEkropo1: You mean exact abc instead of abc.ABC?

NEkropo1: I literally see zero sense in your suggestions xD
Make according to readme... -> created 3 abstract classes.
No! Make according to readme -> inherited from abc -> abc.ABC
No, import ABC from ABC.
Why? For what purpose? Why it bother definition? Have it changes code, or visability, or stile, or anything?
I clearly can't get a thing.

NEkropo1: But I still question why I should do this to both classes inherited from abstract Player.
Explicit over implicit?

wQuelS (approved):
Good

PR #329, opened 2023-02-12
Solution

Zenionoi:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: Player) -> int:
```
usrofgh, line 6: you get a list of Players here, not just one Player's object

app/main.py
```diff
@@ -1 +1,17 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: Elf) -> None:
```
usrofgh, line 10: same here and in others places

usrofgh (changes requested):

YuriiKindrat (approved):
GJ!

PR #332, opened 2023-02-12
Solution DonchykD

DmytryTong:


yura-hudzovskyi (approved):
Good job ;)

app/main.py
```diff
@@ -1 +1,22 @@
...
+from app.players.elves.elf_ranger import ElfRanger
+from app.players.elves.druid import Druid
+from app.players.dwarves.dwarf_warrior import DwarfWarrior
+from app.players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+
+
+def calculate_team_total_rating(list_of_players: list[ElfRanger,
+                                                      Druid,
+                                                      DwarfWarrior,
+                                                      DwarfBlacksmith]) -> int:
```
Nattalli, line 10: you can replace all list types with only one. What do you think what the type of list? Also, change your type

app/main.py
```diff
@@ -1 +1,22 @@
...
+    return sum(player.get_rating() for player in list_of_players)
+
+
+def elves_concert(list_of_elves: list[ElfRanger, Druid]) -> None:
```
Nattalli, line 14: simplify: `list[ElfRanger, Druid]`

app/main.py
```diff
@@ -1 +1,22 @@
...
+    for elf in list_of_elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(list_of_dwarves: list[DwarfWarrior,
+                                               DwarfBlacksmith]) -> None:
```
Nattalli, line 20: simplify: `list[DwarfWarrior, DwarfBlacksmith]`

Nattalli (changes requested):

app/main.py
DmytryTong: I guess I got it. Fixed.

app/main.py
DmytryTong: Done.

app/main.py
DmytryTong: Done.

Nattalli (approved):
GJ!

PR #333, opened 2023-02-13
Solution

ZabFTFT:


app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
...
+
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
+        super().__init__(nickname)
+        self._musical_instrument = musical_instrument
+
+    def play_elf_song(self) -> None:
+        print(
+            f"{self.nickname} is playing a song"
+            f" on the {self._musical_instrument}"
+        )
```
kuznetsb, line 14: class Elf doesn't implement abstract methods. Pycharm should tell you about them.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class Dwarf(Player):
+
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    def eat_favourite_dish(self) -> None:
+        print(f"{self.nickname} is eating {self._favourite_dish}")
```
kuznetsb, line 11: same as for Elf class

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from __future__ import annotations
+
+from app.players.dwarves.dwarf_warrior import DwarfWarrior
+from app.players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+from app.players.elves.druid import Druid
+from app.players.elves.elf_ranger import ElfRanger
+
+
+def calculate_team_total_rating(team: list[Druid | ElfRanger]) -> int:
```
kuznetsb, line 9: I think in annotations it should be list[Player]

app/main.py
```diff
@@ -1 +1,22 @@
...
+    return sum(elf_player_class.get_rating() for elf_player_class in team)
+
+
+def elves_concert(elves: list[ElfRanger | Druid]) -> None:
```
kuznetsb, line 13: list[Elf]

app/main.py
```diff
@@ -1 +1,22 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(
```
kuznetsb, line 18: list[Dwarf]

kuznetsb (changes requested):
Check checklist file. It's written there how correctly inherit another abstract class from abstract class


class Base(ABC):
    @abstractmethod
    def say_hi(self):
        pass


class Derived(Base, ABC):
    pass


class SayHi(Derived):
    def say_hi(self):
        print("Hi!")

app/main.py
ZabFTFT: Thanks for attention to ABC inheritance, I misunderstood description=) Also I made mistake with Player annotation at rating_calculation_function as you pointed. I think other annotations are fine because at readme we are passing lists of children instances not Dwarf and Elf

kuznetsb (approved):
Nice!

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from __future__ import annotations
+
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(elf_player_class.get_rating() for elf_player_class in team)
```
YuriiKindrat, line 9: why do u call variable elf_player_class, if this method could work with every class ?

YuriiKindrat (changes requested):

app/main.py
ZabFTFT: Probably i saw examples in readme and recognized it as a case scenario.So that was in my mind during coding process. Sorry about this kind of mistakes.They are stupid

VladyslavCherkashyn (approved):

Serhii-Leonenko (approved):
Good job!

PR #337, opened 2023-02-13
Solution

viktoria-rybenchuk:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
```
ramsty, line 1: I think we don't need to import ABC here

app/players/elves/elf.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC
```
ramsty, line 1: same here

app/players/dwarves/dwarf.py
viktoria-rybenchuk: Why not? I implement abstract class and import module for it

YuriiKindrat (approved):

PR #344, opened 2023-02-14
Solution

ramsty:


NazarHladaniuk (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
Nattalli, line 1: better to clarify the type of instances inside of list

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list) -> None:
```
Nattalli, line 5: the same here: clarify the list type

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
Nattalli, line 10: the same here

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Nattalli, line 4: you should inherit from `ABC`

app/players/elves/druid.py
```diff
@@ -0,0 +1,18 @@
+from app.players.elves.elf import Elf
+
+
+class Druid(Elf):
+    def __init__(
+            self, nickname: str,
+            musical_instrument: str,
+            favourite_spell: str
+    ) -> None:
+        super(Druid, self).__init__(nickname, musical_instrument)
```
Nattalli, line 10: why did you decide to send `Druid, self` into `super` here? why did you decide to do it only here? please, keep one codestyle everywhere

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Nattalli, line 4: add `ABC` class

Nattalli (changes requested):

app/players/elves/druid.py
ramsty: It was said that best practice not to add it, but Pycharm adds it, so in other cases i deleted it. My bad, didn't see that, corrected

YuriiKindrat (approved):

PR #348, opened 2023-02-14
Solution

Anatolii-Poznyak:


ostapT (approved):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(list_of_player: list[Player]) -> int:
+    return sum(rate.get_rating() for rate in list_of_player)
```
YuriiKindrat, line 7: why do u call variable rate? here will be an instance of Player class inside this variable

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
fix this

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
YuriiKindrat, line 4: the same here

YuriiKindrat (changes requested):

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
+        super().__init__(nickname)
+        self.__musical_instrument = musical_instrument
```
Serhii-Leonenko, line 8: Do you know the difference between protected and private attributes? Fix in all similar cases

Serhii-Leonenko (changes requested):

app/players/elves/elf.py
Anatolii-Poznyak: roger that !

wQuelS (approved):
GJ

PR #354, opened 2023-02-14
Solution

ostapT:


ykopatko (approved):
gj

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,22 @@
...
+    ) -> None:
+        self._skill_level = skill_level
+        super().__init__(
+            nickname=nickname,
+            favourite_dish=favourite_dish
+        )
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
+               f"with skill of the {self._skill_level} level"
```
Serhii-Leonenko, line 19: We don't recommend to use `/`, do the same without it. Fix in all other similar cases

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Serhii-Leonenko, line 4: It should inherit from the ABC also. The same about Dwarf

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #356, opened 2023-02-14
'Solution'

oksanaaam:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elfs: list[Elf]) -> None:
+    for elf in elfs:
+        elf.play_elf_song()
```
ZhAlexR, line 12: it can be done in one line

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
ZhAlexR, line 4: The Dwarf class did not implement any abstract function, but it has to in this case.
But In the task this class has to be abstraction too.
Something like this: 
 ```python
 class Player(ABC):
    @abstractmethod
    def some_method_from_abstract()
         pass
class Dwarf(Player, ABC)
   def_some_method

class DwarfBlacksmith(Dwarf):
    def some_method_from_abstract()
```

File: app/players/elves/elf.py [hunk 3]
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
ZhAlexR, line 4: Same thing with abstraction like in the Dwarf class

ZhAlexR (changes requested):

ZhAlexR (approved):

app/main.py
```diff
@@ -1 +1,15 @@
...
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elfs: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in elfs]
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
Serhii-Leonenko, line 15: List comprehension is the right tool for creating lists, but` for `loop is the right tool for calling functions and computations

Serhii-Leonenko (changes requested):

app/main.py
oksanaaam: > List comprehension is the right tool for creating lists, but`for`loop is the right tool for calling functions and computations

do you mean return code in 2 lines like previous : 
```
for elf in elfs:
        elf.play_elf_song()
        ```

wQuelS (approved):
Good

PR #357, opened 2023-02-14
Solution

NickSavchyn:


dogmoony (approved):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum([member.get_rating() for member in team])
```
amber-marichi, line 7: you can use sum without making it in a list

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elves: list[Elf]) -> list:
+    return [member.play_elf_song() for member in elves]
```
amber-marichi, line 11: according to the task this method should not return anything. just remove the "return" and annotation

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> list:
+    return [member.eat_favourite_dish() for member in dwarves]
```
amber-marichi, line 15: same here. it should not return, just call method on each

amber-marichi (changes requested):
few changes in main file

amber-marichi (approved):
nice! approved

app/main.py
```diff
@@ -1 +1,15 @@
...
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(member.get_rating() for member in team)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    [member.play_elf_song() for member in elves]
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> None:
+    [member.eat_favourite_dish() for member in dwarves]
```
wQuelS, line 15: no need in comprehensions here

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
wQuelS, line 4: image
Do it according to the task description, fix everywhere

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self,
+                 nickname: str,
+                 favourite_dish: str,
+                 skill_level: int) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
+               f"with skill of the {self._skill_level} level"
```
wQuelS, line 14: try to avoid line breaks, fix everywhere

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(member.get_rating() for member in team)
+
+
+def elves_concert(elves: list[Elf]) -> list[None]:
+    return [elf.play_elf_song() for elf in elves]
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> list[None]:
+    return [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
YuriiKindrat, line 15: it is better to use for loops here, cause list comprehension is better to use for creating lists instead of for loop that is better for calling functions, also no need to use return here, it will return a list with None's inside it

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #358, opened 2023-02-14
Solution

inrock1:


app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,19 @@
...
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._hummer_level = hummer_level
+
+    def get_rating(self) -> int:
+        return self._hummer_level + 4
+
+    def player_info(self) -> str:
+        return f"Dwarf warrior {self.nickname}. {self.nickname} " \
+               f"has a hummer of the {self._hummer_level} level"
```
me1nyk, line 19: probably its not a good practice of using line breaks like this (and other player_info methods),
also you have 2 different types of line breaks in this solution
I will approve, but please change them:)
Code seems good for me, nothing complicated, good job

me1nyk (approved):

app/main.py
```diff
@@ -1 +1,10 @@
...
+def calculate_team_total_rating(list_players: list) -> int | float:
+    return sum(player.get_rating() for player in list_players)
+
+
+def elves_concert(list_elf: list) -> None:
+    [elf.play_elf_song() for elf in list_elf]
+
+
+def feast_of_the_dwarves(list_dwarf: list) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in list_dwarf]
```
Serhii-Leonenko, line 10: List comprehension is good for creating lists, but `for` loop is good for calling the functions

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def calculate_team_total_rating(list_players: list) -> int | float:
```
Serhii-Leonenko, line 1: Use more descriptive type annotation. It should looks like `list_players: list[<class name here>]`. The same in all other similar cases

Serhii-Leonenko (changes requested):

app/main.py
inrock1: fixed

app/main.py
inrock1: fixed

app/players/dwarves/dwarf_warrior.py
inrock1: fixed

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from __future__ import annotations
+from app.players.elves.elf_ranger import ElfRanger
+from app.players.elves.druid import Druid
+from app.players.dwarves.dwarf_warrior import DwarfWarrior
+from app.players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+
+
+def calculate_team_total_rating(
+        list_players: list[Druid | ElfRanger]
```
YuriiKindrat, line 9: are u sure about type annotation for list players here? it seems to me that this function is used for calculating Team rating, it can be not only Elves

app/main.py
```diff
@@ -1 +1,23 @@
...
+) -> int | float:
+    return sum(player.get_rating() for player in list_players)
+
+
+def elves_concert(list_elf: list[Druid | ElfRanger]) -> None:
```
YuriiKindrat, line 14: u can use Elf class for type annotation here

app/main.py
```diff
@@ -1 +1,23 @@
...
+    for elf in list_elf:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(
+        list_dwarf: list[DwarfWarrior | DwarfBlacksmith]
```
YuriiKindrat, line 20: the same here

YuriiKindrat (changes requested):

app/main.py
inrock1: Thanks.
I didn't know about possibility to use parent class in type annotation

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from __future__ import annotations
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(
+        list_players: list[Elf | Dwarf]
```
YuriiKindrat, line 7: also u can use class Player here

YuriiKindrat (approved):

PR #365, opened 2023-02-14
Solution

ZhAlexR:


ZabFTFT (approved):
Seems good!

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players_list: list[Elf, Dwarf]) -> int:
+    return sum(player.get_rating() for player in players_list)
+
+
+def elves_concert(elves_list: list[Elf]) -> None:
```
wQuelS, line 9: You can name it just `elves`

app/main.py
```diff
@@ -1 +1,14 @@
...
+    [elf.play_elf_song() for elf in elves_list]
```
wQuelS, line 10: You don't use list which is created by comprehension here, so it is just a memory waste. Fix everywhere

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+
+def feast_of_the_dwarves(dwarves_list: list[Dwarf]) -> None:
```
wQuelS, line 13: `dwarves` ?

wQuelS (changes requested):

wQuelS (approved):
Good

PR #378, opened 2023-02-24
Solution

yhwh6:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from __future__ import annotations
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players_list: list[Player]) -> int:
+    sum_rating = sum([player.get_rating() for player in players_list])
```
ArsenPidhoretskyi, line 8:
```suggestion
    sum_rating = sum(player.get_rating() for player in players_list)
```
Convert to list is not needed

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int
You need to specify type what you need to expect from all your child clases

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
ArsenPidhoretskyi, line 13: It returns str

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #384, opened 2023-02-28
Solution

VladyslavKaminskyi:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: The return type is int. You must specify what type will return all child classes

Fix also in player_info

ArsenPidhoretskyi (changes requested):

PR #386, opened 2023-03-01
Solo

IgoraStar:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. you must specify type which you expect from all your child classes
The same with player_info

ArsenPidhoretskyi (changes requested):

Abnormaltype (approved):

PR #389, opened 2023-03-04
'Solution'

arsenmakovei:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int, you must specify type, which you expect all your child classes will return

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #400, opened 2023-03-24
Solution

maxlapchuk:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    total_rating = 0
+    for player in players:
```
vasylhnatiuk, line 8: try to do it with sum and list comprehension

vasylhnatiuk (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int
You must specify what type must all your child classes return
Fix also in player_info

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #408, opened 2023-03-31
Solution

MaksymUK:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> list:
```
Bohdan-Salamakha, line 10: add additional typehint for return list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1 +1,17 @@
...
+    concert = [elf.play_elf_song() for elf in elves]
+    return concert
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> list:
```
Bohdan-Salamakha, line 15: add additional typehint for return list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

Bohdan-Salamakha (changes requested):
v1

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
...
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self.__favourite_dish = favourite_dish
+
+    def eat_favourite_dish(self) -> str:
```
Bohdan-Salamakha, line 11: return type is None

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
...
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
+        super().__init__(nickname)
+        self.__musical_instrument = musical_instrument
+
+    def play_elf_song(self) -> str:
```
Bohdan-Salamakha, line 11: return type is None

Bohdan-Salamakha (changes requested):
v2

app/players/dwarves/dwarf.py
MaksymUK: If I change to None here, then I have warning in main.py - Expected type 'list[str]', got 'list[None]' instead
I am puzzled how to fix it

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 10: It returns int. You must specify which type you expect from all your child classes
Fix the same problem with player_info

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #409, opened 2023-04-02
'Solution'

callogan:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
Bohdan-Salamakha, line 1: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list) -> None:
```
Bohdan-Salamakha, line 5: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #412, opened 2023-04-08
Solution

AngelinaRemark:


app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,17 @@
...
+class DwarfBlacksmith(Dwarf):
+    def __init__(self,
+                 nickname: str,
+                 favourite_dish: str,
+                 skill_level: int) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
```
vasylhnatiuk, line 13: use () for long str here and all of the code

vasylhnatiuk (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It return int. You must specify which type you expect from all your child classes

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #414, opened 2023-04-10
Task

n1ckmel:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
ArsenPidhoretskyi, line 1: accepts list of what?
Also add inner type annotations to function elves_concert, feast_of_the_dwarves

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. You must specify type which all child classes must return 

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #416, opened 2023-04-11
Solution

Fawnygod:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. You must specify type which all your child classes will return

ArsenPidhoretskyi: The same with player_info

ArsenPidhoretskyi (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> int:
```
ArsenPidhoretskyi, line 13: It returns not int. Please check due to your code. Which type return child classes?

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #432, opened 2023-04-27
'Solution'

PetroHavrysh:


app/main.py
```diff
@@ -1 +1,133 @@
-# write your code here
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. You must specify type which you expect from all your child classes

ArsenPidhoretskyi: The same problem with method player_info

tests/test_main.py
```diff
@@ -6,13 +6,13 @@
...
-from app.players.elves.elf import Elf
-from app.players.elves.elf_ranger import ElfRanger
-from app.players.player import Player
+from app.main import Dwarf
+from app.main import DwarfBlacksmith
+from app.main import DwarfWarrior
+from app.main import Druid
+from app.main import Elf
+from app.main import ElfRanger
+from app.main import Player
```
ArsenPidhoretskyi, line 15: Restore this changes!!!

ArsenPidhoretskyi (changes requested):
image
Not implemented

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+            self,
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
+    ) -> None:
+        super().__init__(nickname=nickname, favourite_dish=favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} with skill of the " \
```
vasylhnatiuk, line 15: use () for long str instead of / here and all of the code

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #443, opened 2023-05-10
Py Elves and Dwarves

alexgubarev88:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Fix the same problem with player_info

app/players/elves/elf_ranger.py
```diff
@@ -0,0 +1,16 @@
...
+class ElfRanger(Elf):
+    def __init__(self, nickname: str,
+                 musical_instrument: str,
+                 bow_level: int) -> None:
+        super().__init__(nickname, musical_instrument)
+        self._bow_level = bow_level
+
+    def player_info(self) -> str:
+        return f"Elf ranger {self.nickname}. " \
+               f"{self.nickname} has bow of the {self._bow_level} level"
```
ArsenPidhoretskyi, line 13:
```suggestion
        return (
            f"Elf ranger {self.nickname}. "
            f"{self.nickname} has bow of the {self._bow_level} level"
        )
```
Better to use brackets instead of \
Format the same in all places

ArsenPidhoretskyi, line 13:
```suggestion
        return (
            f"Elf ranger {self.nickname}. "
            f"{self.nickname} has bow of the {self._bow_level} level"
        )
```
Better to use brackets instead of \
Format the same in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #456, opened 2023-05-23
Solution

AlyonkaB:


app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,18 @@
...
+                 favourite_dish: str,
+                 skill_level: int) -> None:
+        super().__init__(nickname=nickname,
+                         favourite_dish=favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith" \
+               f" {self.nickname} with skill of the" \
+               f" {self._skill_level} level"
```
ArsenPidhoretskyi, line 15:
```suggestion
        return (
            f"Dwarf blacksmith"
            f" {self.nickname} with skill of the"
            f" {self._skill_level} level"
        )
```
Use brackets instead of `\`, fix in all places

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #461, opened 2023-05-25
Elves and Dwarves task complete

thorium-zero:


app/players/player.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        super().__init__()
```
YuriiKindrat, line 6: why do u need this ?

YuriiKindrat (changes requested):

PR #462, opened 2023-05-26
solution

uzlss:


app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
```
vsmutok, line 11: It accepts int. You must specify type which you expect from all child classes
The same in player_info

vsmutok (changes requested):

app/players/player.py
uzlss: accepts? does not it returns int so?

vsmutok: so returns 😁

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
vsmutok, line 14:
```suggestion
    def player_info(self) -> str:

```

vsmutok (changes requested):

app/players/player.py
uzlss: oh im retard, i thought it only printed, without return

vsmutok (approved):
Good 👍

Abnormaltype (approved):

PR #464, opened 2023-05-30
Solution

gleblitvinenko:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players_list: list) -> int:
```
ArsenPidhoretskyi, line 5: list of what?
Also rename players_list to players
no need to add suffix _list
Fix in all places

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Fix the same with player_info

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #465, opened 2023-05-30
implemented elves and dwarves game

Oknametra:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. You must specify type which return all child classes
Fix the same with player_info

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    team_total_rating = sum([player.get_rating() for player in players])
```
ArsenPidhoretskyi, line 7:
```suggestion
    team_total_rating = sum(player.get_rating() for player in players)
```
No need to convert to list)

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> int:
```
ArsenPidhoretskyi, line 13: it returns str, please check your child classes

ArsenPidhoretskyi (changes requested):

app/main.py
Oknametra: sry, my fault

app/players/player.py
Oknametra: yup yup

vasylhnatiuk (approved):

PR #475, opened 2023-06-05
Solution

DmytroTsehelnyk:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Dwarf(Player):
+    def __init__(
+            self,
+            nickname: str,
+            favourite_dish: str
+    ) -> None:
+        super().__init__(nickname)
+        self.nickname = nickname
```
YuriiKindrat, line 11: why do u need this ?

app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,24 @@
...
+class DwarfWarrior(Dwarf):
+    def __init__(
+            self,
+            nickname: str,
+            favourite_dish: str,
+            hummer_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self.nickname = nickname
+        self._favourite_dish = favourite_dish
```
YuriiKindrat, line 13: same, fix this for all classes
why do u need to recreate attributes from parent class in child ?

YuriiKindrat (changes requested):

PR #488, opened 2023-06-06
my_commit

Mr-Furman:


app/main.py
```diff
@@ -1 +1,16 @@
...
+
+    return sum(hero.get_rating() for hero in team)
+
+
+def elves_concert(team: list[Elf]) -> None:
+    [hero.play_elf_song() for hero in team]
+
+
+def feast_of_the_dwarves(team: list[Dwarf]) -> None:
+    [hero.eat_favourite_dish() for hero in team]
```
YuriiKindrat, line 16: it is better to use for loop here, list comprehension is better to use for list creation )

YuriiKindrat (changes requested):

app/main.py
Mr-Furman: I corrected it)

YuriiKindrat (approved):

PR #499, opened 2023-06-09
'Solution'

MarinaKhrunina:


ArsenPidhoretskyi (changes requested):
Please push all files

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> Any:
```
ArsenPidhoretskyi, line 10: Not true, it returns int, check your child classes
the same with player_info
also player_info must be abstractmethods

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #504, opened 2023-06-12
Solution

arturiermolenko:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
olena-hul, line 13: Return annotations are wrong. Should be int and str respectively

olena-hul (changes requested):

PR #514, opened 2023-06-27
Implemented Elves-and-Dwarves task

OleksiiKiva:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> None:
```
Bohdan-Salamakha, line 1: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list) -> None:
```
Bohdan-Salamakha, line 5: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elves: Optional[list]) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: Optional[list]) -> None:
+    for dwarf in dwarves:
+        dwarf.eat_favourite_dish()
```
Bohdan-Salamakha, line 15: not Optional!
indicate what type of variables should be placed into the list like
list[str], list[dict] etc.

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+from app.players.elves.druid import Druid
+
+
+def calculate_team_total_rating(
+        players: list[
+            DwarfWarrior,
+            DwarfBlacksmith,
+            ElfRanger,
+            Druid
+        ]
```
YukihiroSM, line 13: You can use `list[Player]` instead of putting all classes.

app/main.py
```diff
@@ -1 +1,35 @@
...
+) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(
+        elves: list[
+            ElfRanger,
+            Druid
+        ]
```
YukihiroSM, line 22: This will work for all Elfs so you can use `list[Elf]` instead of putting all classes.

app/main.py
```diff
@@ -1 +1,35 @@
...
+) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(
+        dwarves: list[
+            DwarfWarrior,
+            DwarfBlacksmith
+        ]
```
YukihiroSM, line 32: This will work for all Elfs so you can use `list[Dwarf]` instead of putting all classes.

YukihiroSM (changes requested):
Anything else looks good. Improve type hints, please. 
Good job, @OleksiiKiva!

PR #524, opened 2023-07-05
Solution

anastmishchuk:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    total_rating = 0
+    for player in players:
+        total_rating += player.get_rating()
+    return total_rating
```
ArsenPidhoretskyi, line 10: Use sum function

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Fix also player_info

ArsenPidhoretskyi (changes requested):

PR #525, opened 2023-07-05
Solution

bogdandovgan:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
The same fix needed for player_info

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum([player.get_rating() for player in team])
```
ArsenPidhoretskyi, line 7: No need to convert to list
```suggestion
    return sum(player.get_rating() for player in team)
```

app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self,
+                 nickname: str,
+                 favourite_dish: str,
+                 hummer_level: int) -> None:
+        super().__init__(nickname, favourite_dish)
+        self.__hummer_level = hummer_level
+
+    def player_info(self) -> str:
+        return f"Dwarf warrior {self.nickname}. {self.nickname} has " \
+               f"a hummer of the {self.__hummer_level} level"
```
ArsenPidhoretskyi, line 14: Preferred to use parenthesis instead of \
```suggestion
        return (
            f"Dwarf warrior {self.nickname}. {self.nickname} has "
            f"a hummer of the {self.__hummer_level} level"
        )   
```
Fix in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #528, opened 2023-07-10
Solution

3lobites:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
+        pass
```
olena-hul, line 14: Return annotations are wrong. Should be int and str respectively

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #529, opened 2023-07-10
resolved

VladyslavShepilov:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from __future__ import annotations
+
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players_list: list[Player]) -> int:
```
ArsenPidhoretskyi, line 8: remove suffix _list from all variables names

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,16 @@
...
+                 skill_level: int) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def get_rating(self) -> int:
+        return self._skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
+               f"with skill of the {self._skill_level} level"
```
ArsenPidhoretskyi, line 16:
```suggestion
        return (
            f"Dwarf blacksmith {self.nickname} "
            f"with skill of the {self._skill_level} level"
        )
```
Better to use brackets instead of \
Fix in all places

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int you must specify type which you expect from all your child classes
The same fix with player_info

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @VladyslavShepilov not fixed

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi, line 8: Remove _list suffix from the variable name. Fix in all places
```suggestion
def calculate_team_total_rating(players: list[Player]) -> int:
```

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good 👍

PR #535, opened 2023-07-19
solution

nschcur:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+
+def calculate_team_total_rating(players: list) -> int:
```
ArsenPidhoretskyi, line 2: players list of what? Provide inner type

app/main.py
```diff
@@ -1 +1,16 @@
...
+    total_rating = 0
+    for player in players:
+        total_rating += player.get_rating()
+    return total_rating
```
ArsenPidhoretskyi, line 6: Use sum() function

app/main.py
ArsenPidhoretskyi: Check in all places

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
ArsenPidhoretskyi, line 6: FYI inherit from ABC is redundant in this case

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from typing import List
+
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: List[Player]) -> int:
+    total_rating = 0
```
ArsenPidhoretskyi, line 9: this line is redundant

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #547, opened 2023-07-30
Solution

OlPlaksa:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def calculate_team_total_rating(persons: list) -> int:
```
ArsenPidhoretskyi, line 1: list of what? Provide inner type

app/main.py
```diff
@@ -1 +1,10 @@
...
+    return sum(person.get_rating() for person in persons)
+
+
+def elves_concert(persons: list) -> list[str]:
```
ArsenPidhoretskyi, line 5: play_elf_song return None

app/main.py
```diff
@@ -1 +1,10 @@
...
+    return [person.play_elf_song() for person in persons]
+
+
+def feast_of_the_dwarves(persons: list) -> list[str]:
```
ArsenPidhoretskyi, line 9: eat_favorite_dish return None

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 10: it returns int. You must specify type which you expect from all child classes

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(persons: list[Player]) -> int:
+    return sum(person.get_rating() for person in persons)
+
+
+def elves_concert(persons: list[Elf]) -> list[None]:
+    return [person.play_elf_song() for person in persons]
```
vasylhnatiuk, line 11: It will be better to not use list comprehension and return None. You can use:

```
    for elf in elves:
        elf.play_elf_song()
```

vasylhnatiuk: the same in feast_of_the_dwarves

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #549, opened 2023-08-01
Solution

AllaKuksa:


ArsenPidhoretskyi (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
ArsenPidhoretskyi, line 14: What about this return type??

ArsenPidhoretskyi (changes requested):

app/players/player.py
AllaKuksa: str type?

ArsenPidhoretskyi (approved):
Well done!

PR #557, opened 2023-08-11
Solution

eduardhabryd:


app/main.py
```diff
@@ -1 +1,27 @@
...
+def calculate_team_total_rating(
+        players: List[Player]
+) -> int:
+    return sum(
+        player.get_rating()
+        for player in players
+    )
+
+
+def elves_concert(elfs: List[Elf]) -> None:
```
Kroll410, line 16: Comprehension means something should be created, use a for loop for actions to be called

Kroll410 (changes requested):

app/main.py
eduardhabryd: > Comprehension means something should be created, use a for loop for actions to be called

Yeah, but it looks so beautiful with comprehension)))

Kroll410 (approved):

PR #560, opened 2023-08-11
'Solution'

lgSilay:


app/main.py
```diff
@@ -1 +1,29 @@
...
+) -> int:
+    return sum(
+        player.get_rating() for player in players
+    )
+
+
+def elves_concert(
+        elfs: List[Elf]
+) -> None:
+    [
```
Kroll410, line 19: Comprehension means something should be created, use for loop instead

Kroll410 (changes requested):

Kroll410 (approved):

PR #569, opened 2023-08-14
Develop

VitaliyZhabskyi:


tests/test_main.py
ArsenPidhoretskyi, line 1: No need to modify this file. Restore all changes
image

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #586, opened 2023-08-15
solution

EvgenMilokumov:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int. youe must specify type which you expect from all child classes

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

app/players/player.py
EvgenMilokumov: done

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,18 @@
+from app.players.dwarves.dwarf import Dwarf
+
+
+class DwarfBlacksmith(Dwarf):
+    def __init__(
+        self, nickname: str, favourite_dish: str, skill_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
```
vasylhnatiuk, line 8:
```suggestion
        super().__init__(nickname=nickname, favourite_dish=favourite_dish)
```

vasylhnatiuk: it will be better to use key arguments

vasylhnatiuk (changes requested):

app/players/dwarves/dwarf_blacksmith.py
EvgenMilokumov: done, thanks

vasylhnatiuk (approved):
Well done

PR #595, opened 2023-08-18
elves and dwarves

StepanShevchuk:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    result = 0
+    for player in players:
+        result += player.get_rating()
+    return result
```
ArsenPidhoretskyi, line 5: Use sum() function

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
ArsenPidhoretskyi, line 1: List of what? Provide inner type

ArsenPidhoretskyi: Check all places

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Good Job 👍

PR #599, opened 2023-08-20
Solution

VasylTurok:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
pavlejviki, line 5: Since Player is already inheriting from ABC, there's no need for Dwarf to inherit from ABC directly.

pavlejviki (approved):
Well done!

PR #601, opened 2023-08-23
solution

Shum430:


app/main.py
```diff
@@ -1 +1,21 @@
...
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(
+        list_of_players: list[Player]
+) -> int:
+    if list_of_players:
+        return sum(
+            [player.get_rating() for player in list_of_players]
```
pavlejviki, line 11: you can use a generator expression instead of a list comprehension. simply remove [ ]

app/main.py
```diff
@@ -1 +1,21 @@
...
+        )
+    return 0
+
+
+def elves_concert(list_of_elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in list_of_elves]
```
pavlejviki, line 17: there is no need to create a list in this case, a simple loop will be better

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+
+def feast_of_the_dwarves(list_of_dwarves: list[Dwarf]) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in list_of_dwarves]
```
pavlejviki, line 21: there is no need to create a list in this case, a simple loop will be better

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #613, opened 2023-09-03
Solution

AndriiYe:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Dimosphen1, line 4: The `Dwarf` class should be abstract as well because it does not implement all abstract methods from the `Player` class.

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: The `Elf` class should be abstract as well because it does not implement all abstract methods from the `Player` class.

Dimosphen1 (changes requested):

AndriiYe:
So the only purpose of making it abstract is to avoid methods overwriting?

pavlejviki (approved):
Well done!

PR #621, opened 2023-09-07
Solution

Manynik:


app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> list[None]:
+    return [elf.play_elf_song() for elf in elves]
```
pavlejviki, line 11: if you don't use the  return value of the method, then the list is unnecessary. You could replace the list comprehension with a simple loop

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(dwarfs: list[Dwarf]) -> list[None]:
+    return [dwarf.eat_favourite_dish() for dwarf in dwarfs]
```
pavlejviki, line 15: if you don't use the return value of the method, then the list is unnecessary. You could replace the list comprehension with a simple loop

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #623, opened 2023-09-10
solution

antonnech2309:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list) -> None:
```
viktoria-rybenchuk, line 5: You could define that you expect to obtain a list of Elf instances here for elves.

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
viktoria-rybenchuk, line 10: same here

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
+    return sum(player.get_rating() for player in team)
```
viktoria-rybenchuk, line 2: You could define that you expect to obtain a list of Player instances here for players.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+def calculate_team_total_rating(team: [Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: [Elf]) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: [Dwarf]) -> None:
```
Dimosphen1, line 15: Although this type annotation is acceptable, it would be preferable to use` list[Dwarf]` to be more explicit here and in other functions as well.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #625, opened 2023-09-11
Solution

Lebwa1337:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(ls_of_team: list) -> int:
```
viktoria-rybenchuk, line 1: adjust typehint here as List[Player]

viktoria-rybenchuk (changes requested):

PR #629, opened 2023-09-13
Solution

SerhiiBudko:
Good day! When I have been checking my code, I have got a lot of E501 error by "flake8 app" test. I would like to ask your advise, how to solve this problem like that? How to reduce the long line of the code? Thanks a lot!


Dimosphen1 (approved):
LGTM, well done! In case there are any uncovered questions left, you could join a Q&A session to discuss them.

PR #633, opened 2023-09-19
Solution

valikdeb:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum([player.get_rating() for player in players])
```
pavlejviki, line 7: no need in list comprehension, you can use generator expression instead. simply remove []

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #637, opened 2023-09-21
Solution

asdadaversa:


app/main.py
```diff
@@ -1 +1,21 @@
...
+from __future__ import annotations
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    raiting = 0
+    for player in team:
+        raiting += player.get_rating()
```
Dimosphen1, line 10: Implement this block of code in one line with the use of `sum`.

Dimosphen1 (changes requested):
One minor change was requested.

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from __future__ import annotations
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum([player.get_rating() for player in team])
```
pavlejviki, line 8: no need to use list comprehension , you can use generator expression instead which is more efficient. simply remove []

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #644, opened 2023-10-03
Completed the task "Elves and Dwarves".

AndriiHomeniuk1:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
ArsenPidhoretskyi, line 1: list of what? provide inner type

app/main.py
```diff
@@ -1 +1,10 @@
...
+    return sum(rating.get_rating() for rating in players)
+
+
+def elves_concert(elves: list) -> None:
```
ArsenPidhoretskyi, line 5: list of what? provide inner type

app/main.py
```diff
@@ -1 +1,10 @@
...
+    [elf.play_elf_song() for elf in elves]
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
ArsenPidhoretskyi, line 9: list of what? provide inner type

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Also, fix player_info

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: Yes, I understood)
Fixed it.

app/main.py
AndriiHomeniuk1: Fixed it

app/players/player.py
AndriiHomeniuk1: I fixed it)

vasylhnatiuk (approved):

PR #662, opened 2023-10-11
Solution

OlehShumov:


DenPrislipskyi (approved):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from typing import List
+
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: List[Player]) -> int:
```
LiudmylaKulinchenko, line 8: From python 3.10 u can use list instead of typing.List for annotations

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
LiudmylaKulinchenko, line 4: Dwarf/Elf classes should also inherit from ABC

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! 🌺

PR #666, opened 2023-10-11
Solution

D1st3f:


RabenkoYevhenii (approved):
GJ

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum([player.get_rating() for player in players])
```
LiudmylaKulinchenko, line 7: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
LiudmylaKulinchenko, line 7: sum can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
D1st3f: Its very strange, i committed fixes but there is only one commit

fsocie7y (approved):
well done ^_^

PR #669, opened 2023-10-11
Solution

sakhaline:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
```
fedorov-s-od, line 5: It seems you should use super().__init__ here. But I can be wrong

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
```
fedorov-s-od, line 5: I think you should use super().init here )

app/players/player.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
```
fedorov-s-od, line 4: Maybe it's better way to create __init__ method here )

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

LiudmylaKulinchenko (changes requested):

app/players/dwarves/dwarf.py
LiudmylaKulinchenko: to make them abstract you should just inherit from ABC

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🌺

PR #672, opened 2023-10-11
Solution

MiskoRuslan:


Andrii-Silenko (approved):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

LiudmylaKulinchenko (changes requested):

app/players/dwarves/dwarf.py
MiskoRuslan: done :)

LiudmylaKulinchenko (approved):
Good job! 🌺

PR #675, opened 2023-10-11
Solution Elves and Dwarves (DnD)

Oxbay:
I love DnD 😉

### DIFF
--- a/app/players/dwarves/dwarf.py
+++ b/app/players/dwarves/dwarf.py
@@ -1,7 +1,9 @@
+from abc import ABC
+
 from ..player import Player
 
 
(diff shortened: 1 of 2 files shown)


MiskoRuslan (approved):
rly) this solution better than my :_)

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from ..player import Player
+
+
+class Dwarf(Player):
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! 🌷

PR #678, opened 2023-10-11
Solution

Andrii-Silenko:


SelFisher1488 (approved):
can be used as great example!

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #680, opened 2023-10-11
Elves and Dwarves

SelFisher1488:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players_list: list[Player]) -> int:
+    return sum([player.get_rating() for player in players_list])
```
LiudmylaKulinchenko, line 7: sum works with generators so `[]` are not needed

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (changes requested):
All checks have failed

fsocie7y (approved):
well done ^_^

PR #687, opened 2023-10-15
'Solution'

oleksandr-pupchenko:


app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
LiudmylaKulinchenko, line 14: you should add correct annotaions to abstract methods. If you know what type you want to return in classes that will inherit from this one, you should provide accurate annotations already here

LiudmylaKulinchenko (changes requested):
Well done!

LiudmylaKulinchenko (approved):
Great! 🌼

PR #701, opened 2023-10-23
Solution

IvanMozhar:


app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(list_of_players: list[Player]) -> int:
+    return sum(player.get_rating() for player in list_of_players)
+
+
+def elves_concert(list_of_elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in list_of_elves]
```
pavlejviki, line 11: no need to use list comprehension as you don't need to save the result of play_elf_song(). a simple loop is better

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(list_of_dwarves: list[Dwarf]) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in list_of_dwarves]
```
pavlejviki, line 15: see comment above

pavlejviki (changes requested):

pavlejviki (approved):
GJ!

PR #705, opened 2023-10-24
Solution

Ded31:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
YukihiroSM, line 4: You need to inherit from Player, ABC here, because if you are inheriting from Player only - you are required to implement  abstract methods here

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
YukihiroSM, line 4: You need to inherit from Player, ABC here, because if you are inheriting from Player only - you are required to implement  abstract methods here

YukihiroSM (changes requested):

NikOneZ1 (changes requested):
It is a good practice to use __init__.py  to control and organize the package's namespace. It defines what gets imported, ensuring a clean and well-structured public API for the package. Also with __init__.py, you can simplify module imports within the package. Instead of using lengthy import paths, you can import modules directly from the package, making your code cleaner and more readable.
Everything else looks good 👍

Ded31:
I have problems with flake8 because of the init.py file. It's ok?

NikOneZ1 (changes requested):
Sorry, I missed this point with the flake8 check. Please return it the way it was, then everything will be fine!

Ded31:
OK?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
EdAlekseiev, line 1: You should specify inner type of list

EdAlekseiev, line 1:
```suggestion
def calculate_team_total_rating(players: list[Player]) -> int:
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list) -> None:
```
EdAlekseiev, line 5: The same comment as above

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
EdAlekseiev, line 10: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #707, opened 2023-10-29
Develop

DenVia:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+
+
+def calculate_team_total_rating(team: list) -> int:
+    return sum([teammate.get_rating() for teammate in team])
```
LiudmylaKulinchenko, line 4: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job 🦋

PR #715, opened 2023-10-31
Solution

CrazyHilly:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    total_rating = 0
+    for player in players:
+        total_rating += player.get_rating()
+    return total_rating
```
EdAlekseiev, line 10: You can rewrite this function to one line with using `sum` function

EdAlekseiev (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: it returns int, you must specify type which you expect from all child classes

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #724, opened 2023-11-10
Solution

VoloIvanchuk:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
ArsenPidhoretskyi, line 9: It returns int, you should specify a type that returns all your subclasses

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #752, opened 2023-12-03
Solution

vpanchenkoSPS:


app/players/elves/druid.py
```diff
@@ -0,0 +1,19 @@
...
+            favourite_spell: str
+    ) -> None:
+        super().__init__(nickname, musical_instrument)
+        self._favourite_spell = favourite_spell
+
+    def get_rating(self) -> int:
+        return len(self._favourite_spell)
+
+    def player_info(self) -> str:
+        return f"Druid {self.nickname}. {self.nickname} has a "\
```
fsocie7y, line 18: try write ur code without line brackers use ( ) instead if u need

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
fsocie7y, line 4: Elf/Dwarf classes should also be abstract

fsocie7y (changes requested):

PR #759, opened 2023-12-08
Solution

gertsovenko:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(list_of_players: list[Player]) -> int:
+    return sum([player.get_rating() for player in list_of_players])
```
pavlejviki, line 7: list comprehension is redundant. use a generator expression. simply remove the []

pavlejviki (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,14 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
viktoria-rybenchuk, line 4: `Dwarf` and `Elf` should be abstract class as well

viktoria-rybenchuk (changes requested):

PR #764, opened 2023-12-11
Dev

job-artem:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Dimosphen1, line 4: The `Dwarf` class should be abstract since it does not implement all abstract methods from the `Player` class.

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: The `Elf` class should be abstract since it does not implement all abstract methods from the `Player` class.

Dimosphen1 (changes requested):

app/players/dwarves/dwarf.py
job-artem: Why this not fixed ? Is Dwarf is not abstract class ? Since it has abstract method  play_elf_song and passes test "test_main.py::test_classes_should_be_abstract[Dwarf] PASSED              [ 26%]"

image

app/players/elves/elf.py
job-artem: Why this not fixed ? Is Elf is not abstract class ? Since it has abstract method play_elf_song and passes test "test_main.py::test_classes_should_be_abstract[Elf] PASSED                [ 23%]"
image

job-artem: Using provided definition from lesson - i can't get what isn't right!
image

PR #769, opened 2023-12-20
Solution

Artkom98:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list | Player) -> int:
+    total = 0
+    for player in players:
+        total += player.get_rating()
+    return total
```
YukihiroSM, line 10: if players will be type of Player - you will get exception that Player is not iterable. It should be better list[Player]

YukihiroSM: Same for other functions

YukihiroSM (changes requested):
Else looks good to me!

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import abstractmethod, ABC
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
```suggestion
    def get_rating(self) ->  int:

```

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
```suggestion
    def player_info(self) -> str:

```

vasylhnatiuk (approved):

PR #785, opened 2024-01-01
Solution

angelzagniy:


app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def elves_concert(elves: list[elf.Elf]) -> None:
+    for each_elf in elves:
+        each_elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list[dwarf.Dwarf]) -> None:
+    for each_dwarf in dwarves:
+        each_dwarf.eat_favourite_dish()
```
Dimosphen1, line 17: Although this approach of importing files and retrieving classes from them is acceptable, it would be preferable to import classes directly in this task. Please do so. In case you face errors in annotations, you could use `from __future__ import annotations`.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players import player
```
Dimosphen1, line 2: Separate different groups of imports with one blank line according to PEP-8 here and in the `app/players/elves/elf.py`.

Dimosphen1 (changes requested):

PR #800, opened 2024-01-10
Solution

halyna-baklanova:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+from app.players.player import Player
+from typing import List
```
pavlejviki, line 4: typing is part of the standard library, so it comes first. Then, the local application imports

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
```
pavlejviki, line 1: good practice is to add a blank line between each import group

pavlejviki (approved):
Well done, just a minor advice

PR #808, opened 2024-01-16
Solution

yulias97:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
isikach, line 6: You can leave only Player here
(abstract class must have at least one abstract method, class Dwarf - not)

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
isikach, line 6: You can leave only Player here
(abstract class must have at least one abstract method, class Elf - not)

isikach (approved):

app/players/elves/elf.py
NEkropo1: I like that styling, since our Elf is pretty abstract class and you explicitly mention this )

NEkropo1 (approved):
LGTM!

PR #809, opened 2024-01-16
Solution

isikach:


PythonZem (approved):
Your code doesn't pass the tests

Make Dwarfs and Elfs abstract classes and you won't have to initialize abstract methods in them

PR #812, opened 2024-01-17
Solution

oksankakot:


app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elfs: list[Elf]) -> list:
+    return [elf.play_elf_song() for elf in elfs]
```
mdubyna, line 11: we don't need to return anything

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> list:
+    return [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
mdubyna, line 15: we don't need to return anything

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from app.players.player import (Player)
```
mdubyna, line 1: there are no parentheses in other modules, but here there are

mdubyna (approved):
some small mistakes

app/players/elves/elf.py
oksankakot: But we use parental nickname in these class

app/main.py
oksankakot: Thanks

app/main.py
oksankakot: thanks

app/players/elves/elf.py
oksankakot: Understood you, fixed in last commit, thanks

app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elfs: list[Elf]) -> list:
+    [elf.play_elf_song() for elf in elfs]
```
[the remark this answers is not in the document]
oksankakot: Fixed into loop

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> list:
+    [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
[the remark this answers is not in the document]
oksankakot: Fixed

PR #832, opened 2024-01-23
solution

eugenevsn:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
EdAlekseiev, line 1: Specify inner type of list.
```suggestion
def calculate_team_total_rating(team: list[Player]) -> int:
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list) -> None:
```
EdAlekseiev, line 5: The comment as above

app/main.py
```diff
@@ -1 +1,12 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
EdAlekseiev, line 10: The same comment as above

EdAlekseiev (changes requested):

NikOneZ1 (approved):

PR #853, opened 2024-02-04
Solution

fposlushnyi:
5.12. OOP Encapsulation Polymorphism Abstraction

### DIFF
--- a/app/players/dwarves/dwarf.py
+++ b/app/players/dwarves/dwarf.py
@@ -1,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    def eat_favourite_dish(self) -> None:
+        print(f"{self.nickname} is eating {self._favourite_dish}")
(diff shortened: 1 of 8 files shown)


app/players/elves/elf.py
```diff
@@ -0,0 +1,15 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: `Elf` class should inherit from the `ABC` class as well, since it does not implement all abstract methods of a `Player` class, thus being also abstract.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Dimosphen1, line 4: `Dwarf` class should inherit from the `ABC` class as well, since it does not implement all abstract methods of a `Player` class, thus being also abstract.

Dimosphen1 (changes requested):
Several minor changes were requested.

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
```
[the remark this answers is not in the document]
fposlushnyi: ok

PR #859, opened 2024-02-08
Elves and dwarves: solution

ghost:


dmytro-yefimenko:
I just thought that I could write the following parts of code in one line:

`_ = [elf.play_elf_song() for elf in elves]`

`_ = [dwarf.eat_favourite_dish() for dwarf in dwarves]`

YukihiroSM (approved):
That's ok, you are not required to use list comprehension here. Also, in that case you will create another additional (not needed) list. 

The code looks good!

PR #861, opened 2024-02-08
Solution

MaksymNikulyak:


app/main.py
```diff
@@ -1 +1,20 @@
...
+
+def elves_concert(elves: List[Elf]) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarfs: List[Dwarf]) -> None:
+    meals = (Dwarf.eat_favourite_dish(dwarf) for dwarf in dwarfs)
+    for _ in meals:
+        pass
```
EdAlekseiev, line 20: For what?

EdAlekseiev (changes requested):

app/main.py
MaksymNikulyak: If your question is conceptual:
As long as I have demonstrated a good-style way of coding this kind of task, just above in the `elves_concert()` function, I have presented this (admittedly unnecessarily complicated) other way of solving the same task, just as an exhibition of mastering the OOP material (notice the `Dwarf.eat_favourite_dish(dwarf)` instead of the simple `dwarf.eat_favourite_dish()`) and the functions-related part of Python whereas functions are the "first-class" citizens thereof.
Otherwise if you meant to ask from a purely technical viewpoint:
in line 19, every one of the instances of the `.eat_favourite_dish()` method, gets called, which gets us what we want from the `elves_concert()` function.

Will replace the piece with a good-practices code.

EdAlekseiev (approved):
Good job

PR #894, opened 2024-02-28
'Solution'

KravchenkoSerhiii:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from typing import List
```
Dimosphen1, line 1: Add a blank line between different groups of imports according to PEP-8

app/main.py
```diff
@@ -1 +1,17 @@
...
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: List[Elf]) -> None:
+    [elf.play_elf_song() for elf in elves]
+
+
+def feast_of_the_dwarves(dwarves: List[Dwarf]) -> None:
+    for dwarf in dwarves:
+        print(f"{dwarf.nickname} is eating {dwarf._favourite_dish}")
```
Dimosphen1, line 17: Call `eat_favourite_dish` here and make a list comprehension to make code similar to `elves_concert` and preserve the code style

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: Inherit `Elf` from the `ABC` as well to indicate that this class is abstract since it does not implement all abstract methods from the `Player` class

Dimosphen1 (changes requested):

PR #899, opened 2024-03-01
'Solution'

Meyertal:


NikOneZ1 (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
EdAlekseiev, line 1: Specify inner type of list
```suggestion
def calculate_team_total_rating(team: list[Player]) -> int:
```

app/main.py
```diff
@@ -1 +1,13 @@
...
+    all_rating = [i.get_rating() for i in team]
+    return sum(all_rating)
+
+
+def elves_concert(elves: list) -> None:
```
EdAlekseiev, line 6: The same comment as above

app/main.py
```diff
@@ -1 +1,13 @@
...
+    for i in elves:
+        i.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarwes: list) -> None:
```
EdAlekseiev, line 11: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #909, opened 2024-03-04
Solution

Sergunshot:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    sum_of_ratings = 0
+    for player in players:
+        sum_of_ratings += player.get_rating()
+    return sum_of_ratings
```
EdAlekseiev, line 10: Rewrite this method to one line with using function `sum`

EdAlekseiev (changes requested):

app/main.py
Sergunshot: Done

PR #910, opened 2024-03-05
Solution

skeletronys:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
```
Dimosphen1, line 4: Add a blank line between different groups of imports to follow PEP-8

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
+    def __init__(self,
+                 nickname: str,
+                 favourite_dish: str) -> None:
```
Dimosphen1, line 7: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change to all places

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

Dimosphen1 (changes requested):

PR #922, opened 2024-03-11
'Solution'

YESosnovska:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
EdAlekseiev, line 1: You should specify inner type of list
```suggestion
def calculate_team_total_rating(players: list[Player]) -> int:
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+    team_sum = 0
+    for player in players:
+        team_sum += player.get_rating()
+    return team_sum
```
EdAlekseiev, line 5: rewrite to one line with using method `sum`

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elfs: list) -> None:
```
EdAlekseiev, line 8: You should specify inner type of list

app/main.py
```diff
@@ -1 +1,15 @@
...
+    for elf in elfs:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
EdAlekseiev, line 13: You should specify inner type of list

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #928, opened 2024-03-15
Solution

Bujhm99:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
Oleksl888, line 1: You can be more verbose with types inside of list

app/main.py
```diff
@@ -1 +1,15 @@
...
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
```
Oleksl888, line 5: this function can be rewritten as a list comprehension for better efficiency

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elves: list) -> None:
+    for elve in elves:
```
Oleksl888, line 9: Correct names would be:
elf -> elves
dwarf -> dwarves

Oleksl888 (changes requested):

app/main.py
Oleksl888: I think you missed this: please include typing for list contents

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: What will be the types that these methods actually return?

Oleksl888 (changes requested):

Oleksl888 (approved):
NIce one!

PR #937, opened 2024-03-22
Solution

DmytroHlazyrin:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
```
panicua, line 7: here's what [YuriiKindrat](https://github.com/YuriiKindrat) said me about comprehensions with functions, idk if it applies to generators also cause they look alike.

NVIDIA_Share_GdqWWBPbTf

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
panicua, line 4: I'm not sure about this one so you double check it and decide.
from the conditions for the task:
NVIDIA_Share_bxUttUgNHJ

what chat GPT says: ```In Python, a class becomes abstract if it directly inherits from an ABC (Abstract Base Class) and implements at least one abstract method defined in that ABC```

So I belive that this class doesn't count as the abstract one because we didn't use any abstract methods from the parent ```Player``` class inside it, and you need to change it ```class Dwarf(Player, ABC):```
Again I'm not pro so double check it.

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
+        super().__init__(nickname)
+        self.musical_instrument = musical_instrument
```
panicua, line 7: You forgot to save musical_instrument in the protected attribute.

```All elves like playing songs so Elf __init__ method should take an additional argument musical_instrument - and stored it in the protected attribute.```

File: app/players/elves/elf.py [hunk 4]
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
panicua, line 4: Same as with the Dwarf class if I got it right and you agree.

panicua (changes requested):

panicua (approved):
GJ

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Oleksl888, line 6: It is not a requirement to include ABC in this case. This class will have the same behavior with or without including it in parent class.

Oleksl888 (approved):

PR #938, opened 2024-03-23
Solution

panicua:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    total = 0
+
+    for player in players:
```
oleksandra-shershen, line 9: Use list comprehension

app/main.py
```diff
@@ -1 +1,22 @@
...
+        total += player.get_rating()
+
+    return total
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    for elf in elves:
```
oleksandra-shershen, line 16: Check prev comment

app/main.py
```diff
@@ -1 +1,22 @@
...
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> None:
+    for dwarf in dwarves:
```
oleksandra-shershen, line 21: Check prev comment

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int | float:
```
oleksandra-shershen, line 9: If you read the `README.md` file, you can see that we will never get `float` as the type of return value. We always get `int`

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int | float:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> int | float:
```
oleksandra-shershen, line 13: The return type specified for the `player_info` method is incorrect. It should return a string.
Check `README.md` file

oleksandra-shershen (changes requested):
Check my comments
Especially appreciated how you moved the lines))
GJ!

app/main.py
panicua: I can agree with you here ```sum(player.get_rating() for player in players)```. Cause it's not the lsit comprehension, it's more about generators.

app/main.py
panicua: Here's a comment of a mentor - 
NVIDIA_Share_c7YnjcV1c6

Here's a comment from chatGPT when I asked about list comprehension and sum, says sum is good, but comprehension may lead to bad readability

NVIDIA_Share_oqkuOBD1of

app/main.py
panicua: same as previous

app/players/player.py
panicua: since there are no division in those formulas, I agree

app/players/player.py
panicua: since there are no division in those formulas, I agree (2)

oleksandra-shershen (approved):
LGTM!

app/main.py
Oleksl888: basically if we use list comprehension here we will create a list full of None values that we will never use, so for loop is just fine :)

Oleksl888 (approved):
Well done!

PR #941, opened 2024-03-24
Solution for task elves and dwarves

oleksandra-shershen:


vitalinamal (approved):
For me looks good, but PyCharm Illuminates the classes Dwarf and Elf.
It says "Class must implement all abstract methods"

app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in elves]
```
Oleksl888, line 11: Even though list comprehension is faster then a for loop, with this code you are creating a list full of None values that you will never use. Therefor it is better to use for loop

Oleksl888 (approved):
WEll done!

PR #942, opened 2024-03-24
Solution

vitalinamal:


Bujhm99 (approved):
Ok

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Oleksl888, line 5: Pycharm does recommendation about this, however it is not a requirement to mark this class as an ABC. In terms of functionality it will behave the same with or without

Oleksl888 (approved):
Good job!

PR #949, opened 2024-03-25
Solution

y-kondrashova:


CharmNiovi (approved):
GJ

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
```
Oleksl888, line 8: what about return types for these methods?

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Oleksl888, line 6: It is not a requirement to explicitly include ABC in this class. It will behave the same with or without this declaration - as an abstract class.

Oleksl888 (changes requested):

YuriiKindrat (approved):
LGTM!

PR #950, opened 2024-03-25
Solution

CharmNiovi:


app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,21 @@
...
+
+class DwarfWarrior(Dwarf):
+    def __init__(
+            self,
+            nickname: str,
+            favourite_dish: str,
+            hummer_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self.hummer_level = hummer_level
```
SashaKisliy, line 12: must be a protected attribute : _hummer_level

app/players/elves/druid.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Druid(Elf):
+    def __init__(
+            self,
+            nickname: str,
+            musical_instrument: str,
+            favourite_spell: str
+    ) -> None:
+        super().__init__(nickname, musical_instrument)
+        self.favourite_spell = favourite_spell
```
SashaKisliy, line 12: also protected attribute

app/players/elves/elf_ranger.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ElfRanger(Elf):
+    def __init__(
+            self,
+            nickname: str,
+            musical_instrument: str,
+            bow_level: int
+    ) -> None:
+        super().__init__(nickname, musical_instrument)
+        self.bow_level = bow_level
```
SashaKisliy, line 12: _bow_level - protected attribute

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,21 @@
+from app.players.dwarves.dwarf import Dwarf
+
+
+class DwarfBlacksmith(Dwarf):
+    def __init__(
+            self,
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
```
SashaKisliy, line 9: protected attribute

SashaKisliy (approved):
Everything is fine, just change the attributes to 'protected' where I left comments in the code.

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import abstractmethod, ABC
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
```
Oleksl888, line 8: What are the return types for these methods?

app/players/elves/elf.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
Oleksl888, line 6: It is not a requirement to explicitly include ABC in this class. It will behave the same with or without this declaration - as an abstract class.

Oleksl888 (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+    def get_rating(self) -> None:
+        pass
```
Oleksl888, line 10: Leaving None is misleading here. You can put here expected types that will be returned in child instances, or `Any` type in case if different childs can have different return types in same methods.

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #952, opened 2024-03-25
SOLUTION

Nyxan:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> None:
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
+
+
+def elves_concert(elves: list) -> None:
```
DmytroHlazyrin, line 8: It would be better if you specify the type annotation for the list items, because this function can only work for objects that have the "play_elf_song" method.
Do the same for Dwarfs

DmytroHlazyrin (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
```
Oleksl888, line 10: you can write this in one line of code

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: what will be actual return types for these methods?

Oleksl888 (changes requested):

YuriiKindrat (approved):
LGTM!

PR #955, opened 2024-03-26
solution v.1

alonasorochynska:


yossosh (changes requested):
I see a lot of flake8 errors. You need to fix it :)

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team_list: list[Player]) -> int:
+    return sum([player.get_rating() for player in team_list])
```
YuriiKindrat, line 7: no need to use [] with sum, sum could work with generators

YuriiKindrat (approved):

PR #971, opened 2024-03-30
solved

serhii-chaban:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
```
vsmutok, line 6:
```suggestion
    return sum(player.get_rating() for player in players)
```
no need to create list

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list) -> int:
```
Dimosphen1, line 5: Indicate a list of what instances you expect to obtain for each argument in the type annotation. Apply this change to other functions as well

app/main.py
```diff
@@ -1 +1,18 @@
...
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(players: list) -> None:
+    for player in players:
+        if issubclass(player.__class__, Elf):
```
Dimosphen1, line 11: There is no need to call for dunder method here, in case you would like to check the type use the following syntax instead
```suggestion
        if isinstance(player, Elf):
```

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #972, opened 2024-03-30
Solution

Kate-Kotushenko:


app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import abstractmethod, ABC
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
```suggestion
    def get_rating(self) -> int:
```

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
```suggestion
    def player_info(self) -> str:
```

vasylhnatiuk (changes requested):

app/players/player.py
Kate-Kotushenko: Done

app/players/player.py
Kate-Kotushenko: Done

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from __future__ import annotations
```
vsmutok, line 1: I think this import is redundant here

vsmutok (changes requested):

app/main.py
vsmutok, line 1: I thin this import redundant here

vsmutok (changes requested):

app/main.py
Kate-Kotushenko: I have the following error without this import

app/main.py:7: in
E TypeError: 'type' object is not subscriptable

I can avoid using this import if I use the type annotation that looks like this
(team: list)

vasylhnatiuk (approved):

PR #975, opened 2024-03-31
'Solution'

afaidiuk:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
```
Gilgumish, line 1: You must have forgotten push the module Player

Gilgumish (commented):
The logic looks good, but add a module People

Gilgumish (approved):

app/players/dwarves/dwarf.py
YuriiKindrat: @Gilgumish is right, u need to add player module to ur PR

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
```
YuriiKindrat, line 5: could u do this in one line ?)

YuriiKindrat (changes requested):
all tests should pass

PR #976, opened 2024-04-01
solution

RomanDanyl:


vsmutok (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
pavlopro, line 5: There is no need to inherit from ABC

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
pavlopro, line 5: There is no need to inherit from ABC

pavlopro (changes requested):

vasylhnatiuk (approved):

PR #980, opened 2024-04-05
Solution

IraMladanovych02:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
pavlopro, line 1: Provide more detailed annotations for the team parameter, specifying it as a list of Player objects.

Do the same for `elves_concert `and `feast_of_the_dwarves` functions

app/main.py
```diff
@@ -1 +1,15 @@
...
+    total_rating = 0
```
pavlopro, line 2: Attempt to perform this calculation in a single line by using a generator expression and the `sum `function

app/main.py
```diff
@@ -1 +1,15 @@
...
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
+
+
+def elves_concert(elves: list) -> str:
```
pavlopro, line 8: Correct the return annotation. The return type should not be 'str'.

app/main.py
```diff
@@ -1 +1,15 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> str:
```
pavlopro, line 13: Correct the return annotation. The return type should not be 'str'.

pavlopro (changes requested):

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

Dimosphen1 (changes requested):

PR #989, opened 2024-04-09
'Solution'

MarianaTykha:


app/main.py
```diff
@@ -1 +1,15 @@
...
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in elves]
+
+
+def feast_of_the_dwarves(dwarves: list[Dwarf]) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in dwarves]
```
vsmutok, line 15: Do not use List comprehensions if you do not intend to store the result in a variable. Use a regular for loop here.

vsmutok (changes requested):

PR #995, opened 2024-04-17
Develop

AndrewChugreev:


app/main.py
```diff
@@ -1 +1,28 @@
...
+from __future__ import annotations
+from app.players.dwarves.dwarf_warrior import DwarfWarrior
+from app.players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+from app.players.elves.druid import Druid
+from app.players.elves.elf_ranger import ElfRanger
+
+
+def calculate_team_total_rating(
+        team: list
+        [Druid | DwarfBlacksmith | DwarfWarrior | ElfRanger]
```
EdAlekseiev, line 10: You can use Player like inner type of list, because it is parent class Druid, DwarfBlacksmith, DwarfWarrior, ElfRanger and this function uses only `get_rating`, that is specified in Player
```suggestion
        [Player]
```

app/main.py
```diff
@@ -1 +1,28 @@
...
+) -> int:
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
+
+
+def elves_concert(elves: list[ElfRanger | Druid]) -> None:
```
EdAlekseiev, line 18: You can use Elf  like inner type of list
```suggestion
def elves_concert(elves: list[Elf]) -> None:
```

app/main.py
```diff
@@ -1 +1,28 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(
+        dwarves: list
+        [DwarfWarrior | DwarfBlacksmith]
```
EdAlekseiev, line 25: You can use Dwarf like inner type of list
```suggestion
        [Dwarf]
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1002, opened 2024-04-21
Done

Stekloduv:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> float:
```
Dimosphen1, line 1: Indicate a list of what instances you expect to obtain for `players` in the type annotation here and for `elves` and `dwarfs` in the `elves_concert` and `feast_of_the_dwarves` functions correspondingly

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: List[Player]) -> float:
+    return int(sum(player.get_rating() for player in players))
```
Dimosphen1, line 8: There is no need to convert the `sum` output to `int`, leave it as is
```suggestion
    return sum(player.get_rating() for player in players)
```

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: List[Player]) -> float:
```
Dimosphen1, line 7: Indicate that return type could be `float` or `int`in the return type annotation. Use `from __future__ import annotations` import in case you face errors
```suggestion
def calculate_team_total_rating(players: List[Player]) -> float | int:
```

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
```
Dimosphen1, line 1: Add a blank line line here and in the other places to follow PEP-8 import conventions (add a blank line between different groups of imports)

Dimosphen1 (changes requested):

PR #1024, opened 2024-05-19
Complated

Volokitin-Vitaliy:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
+
+
+def calculate_team_total_rating(team: List) -> int:
+    total_rating = 0
+    for player in team:
+        total_rating += player.get_rating()
+    return total_rating
```
Dimosphen1, line 8: Implement this method in one line using `sum` as [mentioned in the checklist](https://github.com/Volokitin-Vitaliy/py-elves-and-dwarves/blob/master/checklist.md#code-efficiency)

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
+
+
+def calculate_team_total_rating(team: List) -> int:
```
Dimosphen1, line 4: Add type annotation to indicate a list of what instances you expect here and in the other methods as well

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1034, opened 2024-05-27
Solution

lilarin:


app/players/elves/elf.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```

DevSpaciX (changes requested):

app/players/elves/elf.py
lilarin: Sure, got it

lilarin: I did it only because my IDE show me an warning: "Class Elf must implement all abstract methods", so i think that would be correct to duplicate ``ABC`` class to pass requirement to impelent abstract functions to further classes

DevSpaciX (approved):

PR #1037, opened 2024-05-28
solution

VShveda:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    for player in elves:
```
DevSpaciX, line 11: Please, use descriptive and correct variable names.
for `elf` in elves: - better

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Elf(Player, ABC):
```

DevSpaciX (changes requested):
Some little mistakes , pls fix them everywhere : )

DevSpaciX (approved):

PR #1040, opened 2024-05-28
solution

AntonKorinchuk:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
DevSpaciX, line 6: Do not repeat abstract methods from a parent if you have an abstract class that inherits another abstract class.

DevSpaciX (changes requested):
Fix that little mistake in all places

DevSpaciX (approved):

PR #1041, opened 2024-05-28
Solution_v1

BileichukIvan:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```

DevSpaciX (changes requested):
Fix that little mistake in all places

DevSpaciX (approved):

PR #1045, opened 2024-05-29
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    counter = 0
+    for player in players:
+        counter += player.get_rating()
+    return counter
```
DevSpaciX, line 5: You can do it in one line using `sum()` func

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def elves_concert(elves: list) -> None:
```
DevSpaciX, line 8: It’s better to write exactly what you expect in the `list`. Please specify the correct typing

app/players/player.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    nickname = []
```
DevSpaciX, line 5: It should be string type , not list ) Implement it inside `__init__` method

app/players/elves/elf.py
```diff
@@ -0,0 +1,15 @@
...
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    def __init__(self,
+                 nickname: str,
+                 musical_instrument: str) -> None:
+        self.nickname = nickname
+        self._musical_instrument = musical_instrument
+        Player.nickname.append(self.nickname)
```
DevSpaciX, line 11: Actually you do not need this )

DevSpaciX (changes requested):

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
DevSpaciX, line 5: Forgotten `ABC` redundat class )) Try always check if you fixed mistake in all places : )

DevSpaciX (changes requested):
Good for now , but fix last little mistake : ))

DevSpaciX (approved):

PR #1046, opened 2024-05-29
Solution

Andrii1000:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(song: list[Elf]) -> None:
```
DevSpaciX, line 10: viriable that stores `list[Elf]` named song ?) Maybe it is better to call it `elves` ?

app/main.py
```diff
@@ -1 +1,17 @@
...
+    for sound in song:
+        sound.play_elf_song()
+
+
+def feast_of_the_dwarves(dish: list[Dwarf]) -> None:
+    for eat in dish:
```
DevSpaciX, line 16: Same with viriable naming : ) Change the `dish` on something more describing

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #1047, opened 2024-05-29
"Solution"

MaglorFeanorson:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```

DevSpaciX (changes requested):
Fix pls little mistake in all places : )

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    @abstractmethod
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
```
DevSpaciX, line 7: Also you do not need here `@abstractmethod` decorator

Oleg-s-s (approved):
👍

PR #1049, opened 2024-05-29
solution

NikitaNikolaitsev:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
+
+
+def elves_concert(elves: list) -> None:
```
DevSpaciX, line 5: It’s better to write exactly what you expect in the `list` . Please specify the correct typing

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```

DevSpaciX (changes requested):
Pls fix some mistakes in all places : )

NikitaNikolaitsev:
I add 2 Question

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
+    # Why don't the tests
+    # throw errors if I haven't
+    # specified the abstract methods get_rating and player_info?
```
DevSpaciX, line 7: Good point , I will check it

app/main.py
```diff
@@ -1 +1,17 @@
...
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
+
+
+def elves_concert(elves: List[Elf]) -> None:  # correct?
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarfs: List[Dwarf]) -> None:  # correct?
```
DevSpaciX, line 15: Right !

DevSpaciX (approved):

PR #1053, opened 2024-05-29
'Solution'

ghost:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
+
+
+def elves_concert(elves: list) -> None:
```
DevSpaciX, line 5: It is better to show what type of data will be in `list ` for exp --> `list[Elf]`

DevSpaciX (changes requested):

Oleg-s-s (approved):
👍

PR #1054, opened 2024-05-30
sol1

Unlie9:


app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```

DevSpaciX (changes requested):

PR #1058, opened 2024-05-31
sol 1

IvanVahanov:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from .players.elves.elf_ranger import ElfRanger
+from .players.elves.druid import Druid
+from .players.dwarves.dwarf_warrior import DwarfWarrior
+from .players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+
+
+usage = [DwarfBlacksmith, DwarfWarrior, Druid, ElfRanger]
```
DevSpaciX, line 7: Why did you make this variable?

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+
+def calculate_team_total_rating(team: list) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list) -> None:
```
DevSpaciX, line 14: it is better to specify exactly what type of data you specify in the list

app/main.py
```diff
@@ -1 +1,21 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
```
DevSpaciX, line 19: Same here : )

DevSpaciX (changes requested):

app/main.py
IvanVahanov: cause flake8 angry at if i not use what i'm imported, but in the task i didn't need to use it

DevSpaciX (approved):

PR #1070, opened 2024-06-09
Solution

frank0190:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from ..player import Player
```
vasylhnatiuk, line 2: please write full path here and everywhere

vasylhnatiuk (changes requested):

vsmutok (approved):
Good Job!

PR #1076, opened 2024-06-19
+

AndriZhok:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def calculate_team_total_rating(teams: list) -> int:
+    result = [sum(team.get_rating() for team in teams)]
```
pavlopro, line 2: Please specify annotation. List of what you expect

Simplify method and use correct naming for variable (player instead of team for example):
`return sum(player.get_rating() for player in teams)`

app/main.py
```diff
@@ -1 +1,13 @@
...
+    return result[0]
+
+
+def elves_concert(elves: list) -> None:
```
pavlopro, line 6: The same about annotation here as well

app/main.py
```diff
@@ -1 +1,13 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarfs: list) -> None:
```
pavlopro, line 11: And here

pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+
+from app.players.player import Player
+
+
+def calculate_team_total_rating(teams: List[Player]) -> int:
+    return sum(player.get_rating() for player in teams)
+
+
+def elves_concert(elves: List[Player]) -> None:
```
pavlopro, line 11: Import and use here `Elf` istead of `Player`

app/main.py
```diff
@@ -1 +1,18 @@
...
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarfs: List[Player]) -> None:
```
pavlopro, line 16: Do the same here

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1081, opened 2024-06-24
Solution

Wallian169:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from __future__ import annotations
+
```
vsmutok, line 2: I think this import redundant here

app/main.py
```diff
@@ -1 +1,18 @@
...
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
```
vsmutok, line 8:
```suggestion
    return sum(player.get_rating() for player in players)
```

vsmutok (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list) -> int:
```
Dimosphen1, line 5: Indicate that you expect to obtain a list of `Player` class instances for `players` in the type annotation

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1103, opened 2024-07-14
Solution

yuliiayedynak:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.players.elves.elf_ranger import ElfRanger
+from app.players.elves.druid import Druid
+from app.players.dwarves.dwarf_warrior import DwarfWarrior
+from app.players.dwarves.dwarf_blacksmith import DwarfBlacksmith
+
+
+def calculate_team_total_rating(
+        players: list[ElfRanger, Druid, DwarfWarrior, DwarfBlacksmith]
```
Dimosphen1, line 8: Better indicate that you expect to obtain here any instances of classes that inherit from `Player` class instead of listing all children of this class

app/main.py
```diff
@@ -1 +1,20 @@
...
+) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elfs: list) -> None:
```
Dimosphen1, line 13: Indicate that you expect to obtain instances of `Elf` class here. Extend type annotation for `feast_of_the_dwarves` as well

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: Since `Elf` does not implement all abstract methods of the `Player` class, `Elf` should also be an abstract class. Inherit it from `ABC` additionally to indicate that. Apply the same change for the `Dwarf` class as well

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1112, opened 2024-07-21
implement solution

HalynaRohatska:
Added packages and files:
players/ 
    player.py
    dwarves/
        dwarf.py
        dwarf_blacksmith.py
        dwarf_warrior.py
    elves/
        elf.py
        druid.py
        elf_ranger.py

### DIFF
--- a/app/players/dwarves/dwarf.py
+++ b/app/players/dwarves/dwarf.py
@@ -1,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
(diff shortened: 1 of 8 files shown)


app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
MNDonut, line 5: No need to inherit from ABC because Player inherits from it

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
MNDonut, line 5: No need to inherit from ABC because Player inherits from it

MNDonut (changes requested):

MNDonut (approved):

PR #1115, opened 2024-07-22
Solution

DSahalatyi:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
MNDonut, line 6: No need to inherit from ABC because Player inherits from it

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return (f"Dwarf blacksmith {self.nickname} "
+                f"with skill of the {self._skill_level} level")
```
MNDonut, line 16: the first and the last brackets should be always on separate lines. fix everywhere

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
MNDonut, line 6: No need to inherit from ABC because Player inherits from it

MNDonut (changes requested):

MNDonut (approved):

PR #1120, opened 2024-07-29
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(players: list[Elf]) -> None:
```
Oleksl888, line 10: we know that we expect a list of elves/dwarves in below functions, so you could use a better name for this and below variables. Players is too general

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
```
Oleksl888, line 8: these methods will return specific types when implemented in child classes, so change annotation according to those

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job sir!

PR #1122, opened 2024-07-29

ValeriiaKyr:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
Oleksl888, line 1: Add type annotation for list elements here and other functions

app/main.py
```diff
@@ -1 +1,12 @@
...
+    return sum(per.get_rating() for per in team)
+
+
+def elves_concert(elves: list) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
+    for dwarv in dwarves:
```
Oleksl888, line 11: `dwarf` is correct singular

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players import player
```
Oleksl888, line 1: it's better to import class explicitly so you don't have to use dot notation below to access it. Check your other imports throughout the project and adjust where necessary

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: these methods will return specific types when implemented in child classes, so change annotation according to those

Oleksl888 (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> str:
+        pass
```
MNDonut, line 10: Are you sure this method returns string everywhere?

MNDonut (changes requested):

oliasymak24 (approved):

PR #1123, opened 2024-07-30
Solution

pashawarganov:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
oliasymak24, line 13: Are you sure that these two functions return everywhere None? these methods will return specific types

oliasymak24 (changes requested):

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+            self,
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} " \
```
Oleksl888, line 15: We do not use `\` for breaking lines as per our styling convention - check throughout the file and change accordingly

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
```
Oleksl888, line 2: It is a good habit to separate different import categories with blank line

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Nice! 💪

PR #1124, opened 2024-07-30
Solution

Ivan-Shakhman:
I hope its correct solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,12 +1,17 @@
-def calculate_team_total_rating(players: list) -> int:
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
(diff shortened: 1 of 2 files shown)


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
oliasymak24, line 13: Are you sure that there two methods return None everuwhere?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
oliasymak24, line 1: Add type annotation for list elements here, check in all places

oliasymak24 (changes requested):

MNDonut (approved):

PR #1126, opened 2024-07-30
Solution Politov

Alexandr-Politov:


app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
LiudmylaKulinchenko, line 4: `each player should be an instance of one of these four classes. But to have better code design, you should implement the abstract classes Elf, Dwarf and Player, where you can put the common logic.` - Elf/Dwarf should also inherit from ABC

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 💪

PR #1128, opened 2024-07-30
first commit

panichqa:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
Oleksl888, line 1: Annotate types in  list contents in each function

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from ..player import Player
```
Oleksl888, line 1: do not use relative imports - is a better practice to use full path e.g. `from app.player...`
check in all files and fix where necessary

app/players/elves/druid.py
```diff
@@ -0,0 +1,19 @@
+from .elf import Elf
+
+
+class Druid(Elf):
+    def __init__(
+        self,
+        nickname: str,
+        musical_instrument: str
+        , favourite_spell: str
```
Oleksl888, line 9: this comma looks misplaced

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: use correct return types here since your methods are not returning None when redefined

Oleksl888 (changes requested):

PR #1130, opened 2024-07-30
Solution for py-elves-and-dwarves

AndriiHamasa:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
```
Oleksl888, line 1: Include type annotation for list contents in each function

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
...
+            self,
+            nickname: str,
+            favourite_dish: str,
+            skill_level: int
+    ) -> None:
+        super().__init__(nickname, favourite_dish)
+        self._skill_level = skill_level
+
+    def player_info(self) -> str:
+        return f"Dwarf blacksmith {self.nickname} with skill of the " \
```
Oleksl888, line 15: As per style guide we do not use `\` to break lines - check in all files and fix where necessary

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: use correct type annotations here

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from typing import List
+
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: List[Player]) -> int:
```
LiudmylaKulinchenko, line 8: In the newest python versions you can use default list instead of typing.List
I believe, you can do it in your case

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
LiudmylaKulinchenko, line 4: According to the task description, Dwarf and Elf classes should also inherit from ABC, check it again

LiudmylaKulinchenko (changes requested):

app/players/dwarves/dwarf.py
LiudmylaKulinchenko: `Thus, each player should be an instance of one of these four classes. But to have better code design, you should implement the abstract classes Elf, Dwarf and Player, where you can put the common logic.`

LiudmylaKulinchenko (approved):
Nice! 💪

PR #1131, opened 2024-07-30
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elf: list[Elf]) -> None:
```
Oleksl888, line 10: elf is not the most suitable name - as a convention in python we use names with plural forms for lists or other collections of data. Or you can add suffix `_list` in this particular case.

app/main.py
```diff
@@ -1 +1,17 @@
...
+    for elfs in elf:
```
Oleksl888, line 11: `elves` is a plural form for `elf`
`dwarves` is a plural form for `dwarf`

app/main.py
```diff
@@ -1 +1,17 @@
...
+        elfs.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarf: list[Dwarf]) -> None:
+    for dwarfs in dwarf:
```
Oleksl888, line 16: change naming here as well to follow the logic explained above

Oleksl888 (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
LiudmylaKulinchenko, line 4: `Thus, each player should be an instance of one of these four classes. But to have better code design, you should implement the abstract classes Elf, Dwarf and Player, where you can put the common logic.` - Elf and Dwarf should also inherit from ABC

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 💪

PR #1132, opened 2024-07-30
Solution

zazmarga:


app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
```suggestion
    def get_rating(self) -> int:

```

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
```suggestion
    def player_info(self) -> str:

```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+def calculate_team_total_rating(team_players: list[Player]) -> int:
+    return sum(each.get_rating() for each in team_players)
+
+
+def elves_concert(team_elves: list[Elf]) -> None:
+    [player.play_elf_song() for player in team_elves]
+
+
+def feast_of_the_dwarves(team_dwarves: list[Dwarf]) -> None:
+    [player.eat_favourite_dish() for player in team_dwarves]
```
vsmutok, line 15: If you don't need to store the result in variable, in this case, it's better to use a regular loop instead of a list comprehension

vsmutok (changes requested):

vsmutok (approved):
Great

PR #1133, opened 2024-07-30
Solution

Bogdan-Rylov:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: these methods will have different return types when redefined in child methods

Oleksl888 (changes requested):

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
LiudmylaKulinchenko, line 4: `Thus, each player should be an instance of one of these four classes. But to have better code design, you should implement the abstract classes Elf, Dwarf and Player, where you can put the common logic.` - Elf and Dwarf should also inherit from ABC

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 🔥

PR #1136, opened 2024-07-31
Develop

Sergi0bbb:


app/main.py
```diff
@@ -1 +1,17 @@
...
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(feast: list[Dwarf]) -> None:
```
Oleksl888, line 15: why would you call a list of dwarves as a `feast`? can it be something more relevant?

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from ..player import Player
```
Oleksl888, line 2: Using relative imports is not as per best practices as it may be tied to location from where the script is being run. Please change where necessary

Oleksl888 (changes requested):

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
+from .dwarf import Dwarf
```
Oleksl888, line 1: relative imports not fixed - check in all files

Oleksl888 (changes requested):

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from app.players.player import Player
```
Oleksl888, line 2: It is also a good practice to leave a blank line between different types of import

Oleksl888 (approved):
Approved. But please mind the last comment about imports - you need to add blank lines in files dwarf.py and elf.py

PR #1138, opened 2024-07-31
Solution

olena4reunited:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self._nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: These methods have different return types in child classes

Oleksl888 (changes requested):

Oleksl888 (approved):
Superstar! :star:

PR #1139, opened 2024-07-31
Solution

mnezbrytska:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import abstractmethod, ABC
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: These methods will have different return types in child classes

Oleksl888 (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,15 @@
+from abc import abstractmethod, ABC
+from typing import Any
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> Any:
```
Oleksl888, line 10: `Any` may be possible but we know for a fact which type is returned in each child class so you can be more specific

app/players/player.py
```diff
@@ -0,0 +1,15 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> Any:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
Oleksl888, line 14: Double check which type is being returned in child classes

Oleksl888 (changes requested):

Oleksl888 (approved):
Great job!

PR #1141, opened 2024-07-31

Maksym-Turenko:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
```
Oleksl888, line 1: You may add type annotation for list contents in all functions

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: these methods will have specific return types when defined in child classes

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from typing import List
```
Oleksl888, line 1: Not a mistake, just reminding that as of python 3.9+ you may use build-in `list` object for type annotations.

Oleksl888 (approved):
Good job! :+1:

PR #1143, opened 2024-07-31
adds classes and methods for Elfs and Dwarfs

egor250330:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+from typing import List
```
Oleksl888, line 4: 1. There is a convention for order of imports that imports from build in libraries must come first. Also you good practice to separate different groups of imports by blank lines
2. You actually can you build-in `list` object for type annotations as of python 3.9

app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,19 @@
+from .dwarf import Dwarf
```
Oleksl888, line 1: relative imports is not a recommended way since imports may or may not work based on the directory from which python script is being run, please change this in all files

app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
Oleksl888, line 9: Check what return types are used in child classes and set accordingly

Oleksl888 (changes requested):

PR #1147, opened 2024-08-06
elves and dwarves

AndreyHMkr:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
pavlopro, line 4: Please inherit from ABC as well

app/players/elves/elf.py
```diff
@@ -0,0 +1,16 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
pavlopro, line 4: Here as well

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1171, opened 2024-08-29
Solution

bdnnk:


app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Dimosphen1, line 4: Since `Elf` does not implement all abstract methods of the `Player` class, `Elf` should also be an abstract class. Inherit it from `ABC` additionally to indicate that. The same relates to the `Dwarf` class as well

Dimosphen1 (changes requested):
Several minor changes were requested

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
```
pavlopro, line 1: Add blank line between different import groups

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC
```
pavlopro, line 1: Add blank line between different import groups

pavlopro (changes requested):

LLkaia (approved):

PR #1190, opened 2024-09-26
Solution

Atikiho:


app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Mxbely, line 4: You inherit the Elf class from the Player abstract class, in the Elf class you still need to inherit from the abstract class, because you don't use Player abstract methods:
```
    @abstractmethod
    def get_rating(self) -> None:
        pass

    @abstractmethod
    def player_info(self) -> None:
        pass
```
for example:
`class Elf(Player, ABC):`

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Mxbely, line 4: You inherit the Elf class from the Player abstract class, in the Elf class you still need to inherit from the abstract class, because you don't use Player abstract methods:
```
    @abstractmethod
    def get_rating(self) -> None:
        pass

    @abstractmethod
    def player_info(self) -> None:
        pass
```
for example:
`class Elf(Player, ABC):`

Mxbely (changes requested):

PR #1201, opened 2024-10-02
Solution

vladyslav-tmf:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
Mxbely, line 4: You also need to add an abstract class to the inherited classes, because, class Dwarf inherited from abstract class People and don't reload `get_rating` and `player_info` methods.

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
Mxbely, line 4: See comment from Dwarf class

Mxbely (changes requested):

Mxbely (approved):

app/players/dwarves/dwarf.py
vladyslav-tmf: > You also need to add an abstract class to the inherited classes, because, class Dwarf inherited from abstract class People and don't reload `get_rating` and `player_info` methods.

That's how I originally wanted to implement it, the Elf and Dwarf classes were highlighted in the IDE, but I wasn't sure for some reason.
I fixed it, thanks!)

AnyoneClown (approved):
LGTM!

PR #1204, opened 2024-10-03

yuliiapyrih:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
LiudmylaKulinchenko, line 4: Elf and Dwarf should also inherit from ABC -- it will show that you don't want to have an ability to create an instance from the class

LiudmylaKulinchenko (changes requested):

app/players/dwarves/dwarf.py
yuliiapyrih: Done

Oleksii-Siviolov (approved):

PR #1207, opened 2024-10-06
 Solution-py-elves-and-dwarves

skyfoxwork:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
94nj111, line 4: Should inherit ABC since you don't reload abstract methods, same for Elf

94nj111 (changes requested):

94nj111 (approved):
Good job!

Nattalli (approved):

PR #1227, opened 2024-10-09
Solution

yulia-pl:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
```
Bondzik-S, line 1: I think it is important to import ABC for Draft here

app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from app.players.player import Player
+
```
Bondzik-S, line 2: And here (ABC) 👆

PR #1228, opened 2024-10-09
Solution

Bondzik-S:


VolodymyrSemchysyn (approved):

app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum(player.get_rating() for player in players)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in elves]
```
AnyoneClown, line 11: You don't need to use List Comprehension here. We are using it, if we want to use this list later. If don't need it - use `for` loop

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def feast_of_the_dwarves(dwarfs: list[Dwarf]) -> None:
+    [dwarf.eat_favourite_dish() for dwarf in dwarfs]
```
AnyoneClown, line 15: Same here

AnyoneClown (changes requested):

AnyoneClown (approved):
GJ!

PR #1232, opened 2024-10-09
Solution

Le0n-K:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.players.player import Player
+from app.players.dwarves.dwarf import Dwarf
+from app.players.elves.elf import Elf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    total_rating = 0
+    for player in players:
+        total_rating += player.get_rating()
+    return total_rating
```
FursykIvan, line 10: You can use refactoring using a generator and the sum function.

return sum(...)

FursykIvan (changes requested):

FursykIvan (approved):

PR #1275, opened 2024-11-18
'Solution'

dottfmar:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    result = 0
```
julia4406, line 2: May be will be better for players type instead the  list-type to use list with instances of Players, smth like List[Players]...

app/main.py
```diff
@@ -1 +1,15 @@
...
+    return result
+
+
+def elves_concert(elves: list) -> None:
+    for elf in elves:
+        elf.play_elf_song()
+
+
+def feast_of_the_dwarves(dwarves: list) -> None:
+    for dwarf in dwarves:
```
julia4406, line 14: list of Dwarf

julia4406 (approved):
You've done a great work!

PR #1288, opened 2024-11-30
Created all packeges, modules, classes and methods

shafran-andriy:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+from app.players.player import Player
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    sum_raiting = 0
```
pavlopro, line 7: simplify to:

pavlopro (changes requested):

app/main.py
shafran-andriy: fixed

pavlopro (approved):
GJ

PR #1290, opened 2024-12-04
Solution

julia4406:


tests/test_main.py
```diff
@@ -14,7 +14,6 @@
 from app.players.elves.elf_ranger import ElfRanger
 from app.players.player import Player
 
-
```
L1nk3rrr, line 17: no need to change something in tests

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+
```
L1nk3rrr, line 1: empty line

L1nk3rrr (changes requested):
Logically all is good, look for your `__init__` methods and their styling, one attribute per new line, and `) -> None`:  must be on the same indentation as the method name

tests/test_main.py
julia4406: This happened accidentally.  When I reordered my project into packages PyCharm changed all imports in my files by itself, and in tests too). I had spent all day to find this and resolve why my code doesn't work. So I just restored imports in test-main.py from origin file again.
All imports in tests looked like:
from app import ElfRanger
from app import Player

L1nk3rrr (approved):
Great work!

PR #1301, opened 2024-12-06
'Solution'

OLypovenko:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from __future__ import annotations
+from typing import List
```
Dimosphen1, line 2: Add a blank line between different import groups to follow PEP-8

app/players/elves/elf_ranger.py
```diff
@@ -0,0 +1,17 @@
+from app.players.elves.elf import Elf
+
+
+class ElfRanger(Elf):
+    def __init__(self,
+                 nickname: str,
+                 musical_instrument: str,
+                 bow_level: int) -> None:
```
Dimosphen1, line 8: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change here and in the other similar cases

app/players/elves/elf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #1308, opened 2024-12-07
Solution

ostroverkhovaa:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(players: list) -> int:
+    return sum([player.get_rating() for player in players])
```
vsmutok, line 2:
```suggestion
    return sum(player.get_rating() for player in players)
```
no need to create  list here

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1315, opened 2024-12-09
Solution

Chukhno-Maksym:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
+        super().__init__(nickname)
+        self._favourite_dish = favourite_dish
+
+    @abstractmethod
```
dottfmar, line 10: The Dwarf class doesn't need to be abstract because it doesn't define any additional abstract methods itself. Instead, it simply implements the eat_favourite_dish method, which is required by its subclasses (DwarfWarrior and DwarfBlacksmith). Since the task specifies that Dwarf will be a concrete class that provides common functionality for its subclasses, we don't need to make Dwarf an abstract base class (ABC). But it also works with ABC as you wrote.

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC, abstractmethod
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
+    def __init__(self, nickname: str, musical_instrument: str) -> None:
+        super().__init__(nickname)
+        self._musical_instrument = musical_instrument
+
+    @abstractmethod
```
dottfmar, line 10: The Elf class doesn't need to be abstract because it doesn't define any additional abstract methods itself. Instead, it simply implements the play_elf_song method, which is required by its subclasses (ElfRanger and Druid). Since the task specifies that Elf will be a concrete class that provides common functionality for its subclasses, we don't need to make Elf an abstract base class (ABC). But it also works with ABC as you wrote.

dottfmar (approved):
Well done!

PR #1318, opened 2024-12-10
Solution

taraskhom:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(players: list[Player]) -> int:
+    return sum([player.get_rating() for player in players])
```
DanilTrukhanov, line 7: you can omit here [] brackets to prevent creating a redundant list

DanilTrukhanov (approved):
good job!

app/main.py
taraskhom: Thanks!

PR #1327, opened 2024-12-10
Elves-and-dwarves_Solution

SummerKatrin:


TretiakHelen (approved):

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
L1nk3rrr, line 5: Why you inherit from ABC again?

L1nk3rrr (changes requested):
Please do all comments from AI, annotation must be right.
Your code is well done and good structured, but there are some points for improvement

PR #1331, opened 2024-12-11
Solution

MindDevastation:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from collections.abc import Iterable
+
+
+def calculate_team_total_rating(team: Iterable) -> int:
```
Dogober, line 4: Here we expected exactly list of Player -> list[Player] if i understand right

Dogober (approved):
Great cod!

PR #1340, opened 2024-12-12
Solution

toku666:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from ..player import Player
```
MaksVakulenko, line 1: it is not a good practice to use: ..player.
better: app.players.dwarves

MaksVakulenko (approved):
see my comment

PR #1341, opened 2024-12-12
solution

MaksVakulenko:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def calculate_team_total_rating(player: list) -> int:
+    return sum([player.get_rating() for player in player])
```
4ikkamony, line 2: you don't need [] inside sum(), it unnecessarily creates a list

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+
+def elves_concert(elf: list) -> None:
+    [el.play_elf_song() for el in elf]
```
4ikkamony, line 6: same thing with [] here, it's better to use for loop to not make a list in memory

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+
+def feast_of_the_dwarves(dwarf: list) -> None:
+    [dwa.eat_favourite_dish() for dwa in dwarf]
```
4ikkamony, line 10: and here, [] not needed

4ikkamony (approved):
Great job!
Only little issues with unneccessary [] here and there

PR #1346, opened 2024-12-13
Solution

bodiakof:


app/players/elves/elf.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
+
+from app.players.player import Player
```
Professor-Douel, line 3: Here too, you can import ABC from Player:
`from app.players.player import Player, ABC`
don`t need the first import

Professor-Douel (changes requested):

L1nk3rrr (approved):

PR #1347, opened 2024-12-13
Solution

Professor-Douel:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list) -> None:
```
Dogober, line 5: It will be better if you add typing anotation to the func's arguments like team: list[Player], etc.

Dogober (approved):
In general seems good !

PR #1424, opened 2025-02-06
Виконав

GGZahar1:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list) -> int:
```
SerhiyTo, line 5: It would be great if you could specify the type of data that the list takes inside it

app/players/player.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
```
SerhiyTo, line 8: You can also use `raise NotImplementedError` instead of the usual `pass`. It's not a criticism, it's just a simple piece of advice

SerhiyTo (approved):

PR #1436, opened 2025-02-14
Solution

Nikita-pyth:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import abstractmethod, ABC
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
```
oliasymak24, line 9: Pay attention which type return these abstract methods in children's classes

app/players/elves/elf.py
```diff
@@ -0,0 +1,14 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
oliasymak24, line 4: Elf/Dwarf classes should also be abstract

oliasymak24 (changes requested):

app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> int:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
oliasymak24, line 13: Here the same situation, can't return None
Try to check all the same situation when I left some comments

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1460, opened 2025-02-25
solution

Anollinna:


Sotnikov-100 (approved):
add import in main.py ----->
from app.players.elves.elf_ranger import ElfRanger
from app.players.elves.druid import Druid
from app.players.dwarves.dwarf_warrior import DwarfWarrior.   

and it is necessary to implement the method  --->
if __name__ == "__main__":

PR #1519, opened 2025-03-27
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
removed commented: abstract methods (get_rating and player_info) from the Dwarf and Elf classes.
_favourite_dish and _musical_instrument are protected

PR #1555, opened 2025-04-29
Added a solution for "Elves and Dwarves"

olitm:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def calculate_team_total_rating(teams: list) -> int:
+    total_rating = 0
+    for player in teams:
+        total_rating += player.get_rating()
+
+    return total_rating
```
vsmutok, line 6: try to use `sum()` and comprehension here

vsmutok (changes requested):

PR #1638, opened 2025-06-25
Solution

DrGregoryHouse4676:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List
```
Arch0998, line 1: Why do you need this import? What does it do?

PR #1656, opened 2025-07-01
Implemented solution

Otst0y:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
```
Artcrafterrra, line 6: In my opinion, we don't need to make the Dwarf class abstract.

app/players/elves/elf.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Elf(Player, ABC):
```
Artcrafterrra, line 6: Same as for Dwarf

Artcrafterrra (approved):
LGTM based on my comments, but I'm not sure.

PR #1658, opened 2025-07-01
solution

Sashoxxx:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
```
dimashevchukk, line 6: can make nickname atr protected

PR #1688, opened 2025-07-13
Solution

Evgeniya0208:


app/main.py
```diff
@@ -1 +1,15 @@
...
+from app.players.player import Player
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    return sum(player.get_rating() for player in team)
+
+
+def elves_concert(elves: list[Elf]) -> None:
+    [elf.play_elf_song() for elf in elves]
+
```
vsmutok, line 12: If you don't need to return a result or store it in a variable, it's better to use a regular loop instead of a list comprehension.

vsmutok (changes requested):

app/main.py
Evgeniya0208: fixed

PR #2018, opened 2026-03-04
Solution

Reverad:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+
+from app.players.player import Player
+
+
+class Dwarf(Player, ABC):
+    def __init__(self, nickname: str, favourite_dish: str) -> None:
```
pawunder1992, line 7: You dont needed  add parents class ABC to Dwarf and Elf(class Dwarf(Player) only) becuse Player is inherited from ABC

pawunder1992 (approved):
delete class ABC from Elf and Dwarf

PR #2071, opened 2026-03-27
update

o7111023-pixel:


app/players/dwarves/dwarf_blacksmith.py
```diff
@@ -0,0 +1,22 @@
+from .dwarf import Dwarf
```
vsmutok: It is better to use absolute imports instead of relative ones.

app/players/dwarves/dwarf_warrior.py
```diff
@@ -0,0 +1,22 @@
+from .dwarf import Dwarf
```
vsmutok: It is better to use absolute imports instead of relative ones.

app/players/elves/elf_ranger.py
```diff
@@ -0,0 +1,22 @@
+from .elf import Elf
```
vsmutok: It is better to use absolute imports instead of relative ones.

app/players/elves/elf.py
```diff
@@ -0,0 +1,18 @@
+from app.players.player import Player  # Імпорт базового класу Player
```
vsmutok: remove this comment

vsmutok (changes requested):

vsmutok (changes requested):
Tests failed. 
Please ensure that your code passes all tests before submitting it for review.

vsmutok (approved):
Good Job!

PR #2086, opened 2026-04-04
Solution

vova-dev-python:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
```
VadimPusiak788: You should separate imports by group and separate them with blank lines

Standard library imports
Related third-party imports
Local application/library imports

app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,11 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Dwarf(Player, ABC):
```
VadimPusiak788: ABC is not really necessary here, because Player already inherits from ABC

VadimPusiak788 (changes requested):

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+from abc import ABC
+
+
+class Elf(Player, ABC):
```
VadimPusiak788: ABC is not really necessary here, because Player already inherits from ABC

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from __future__ import annotations
+from app.players.player import Player
```
VadimPusiak788, line 3: You should separate imports by group and separate them with blank lines

Standard library imports
Related third-party imports
Local application/library imports

VadimPusiak788 (changes requested):

app/players/elves/elf.py
vova-dev-python: Good catch! I've removed it to keep the code cleaner. Thanks!

app/main.py
vova-dev-python: Grouped imports according to PEP 8. Now the structure is compliant with the standards.

VadimPusiak788 (approved):

PR #2087, opened 2026-04-05
solution

PavelPereverzev1:


alisahoys (approved):
You have some wrong return type annotations like here: def player_info(self) -> None: where you clearly return a string. Good abstract class usage!

PR #2099, opened 2026-04-07
Solution

amesemyta1:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from app.players.player import Player
+
+
+class Dwarf(Player):
```
andriihrechko, line 4: Dwarf should be abstract. Use class Dwarf(Player, ABC):

app/players/elves/elf.py
```diff
@@ -0,0 +1,12 @@
+from app.players.player import Player
+
+
+class Elf(Player):
```
andriihrechko, line 4: Elf should be abstract. Use class Elf(Player, ABC):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from typing import List
```
andriihrechko, line 1: You shouldn't use 'List' from typing. Use default built-in list.

andriihrechko (approved):

PR #2207, opened 2026-07-01
Solution

BogdanDemidkin2000:
Solution


hladiivolodymyr (changes requested):
The naming convention is inconsistent. You're mixing protected (_) and private (__) attributes. For example  self._musical_instrument and self.__favourite_spell. Replace the __ prefix with _ to keep the naming consistent.

hladiivolodymyr (approved):

PR #2216, opened 2026-07-06
Solution

Krastidney:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.players.player import Player
+from app.players.elves.elf import Elf
+from app.players.dwarves.dwarf import Dwarf
+
+
+def calculate_team_total_rating(team: list[Player]) -> int:
+    total = 0
+    for player in team:
+        total += player.get_rating()
```
hladiivolodymyr: This works well. Consider using the sum() method here to simplify the implementation.

hladiivolodymyr (changes requested):

vsmutok (approved):
Looks Good!

PR #2259, opened 2026-08-11
Solution DK

DarinaK-k:


app/players/player.py
```diff
@@ -0,0 +1,14 @@
...
+class Player(ABC):
+    def __init__(self, nickname: str) -> None:
+        self.nickname = nickname
+
+    @abstractmethod
+    def get_rating(self) -> None:
+        pass
+
+    @abstractmethod
+    def player_info(self) -> None:
```
Sn1kls, line 13: An abstract method declares the contract, so it should be annotated with what subclasses must return, not `None`.

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def calculate_team_total_rating(team: list) -> int | float:
```
Sn1kls, line 1: Please bring back what you had in the first commit: the imports and the parameterised hints.
You got this right originally – a bare `list` doesn't say what's inside it, and the task examples use `list[Player]`.

Sn1kls (changes requested):

Sn1kls (approved):
Approving, nice work!

PR #2283, opened 2026-08-17
Task: Elves and dwarves

Andrey9417:


BohdanSarai (approved):

Andrey9417:
@mateacademy-ai-mentor get yourself together and stop writing nonsense. all files are there, READ them. all comments are invalid

Andrey9417:
@mateacademy-ai-mentor 

which tests are failing you say? the ones which were all green since first commit? Fixed typo with instruments. As for return types - there is no contract regarding how these methods should look like in Player, they are marked abstract and declared - exactly as was asked in the task. When they are implemented, they receive appropriate return types. And as already mentioned, all test green, at least check the results before posting messages

PR #2311, opened 2026-08-26
My_Solution

Oleksandr-131:


app/players/dwarves/dwarf.py
```diff
@@ -0,0 +1,10 @@
+from ..player import Player
```
vsmutok: It’s better to use absolute imports instead of relative imports.
Please fix this here and in the other modules as well.

vsmutok (changes requested):

